PR #5, opened 2019-11-27
Adding ApplicationNavigation and ApplicationIntlContext concepts

tbiethman:
### Summary
This PR adds a couple new things:

* An implementation of ApplicationNavigation that integrates with the existing application concepts (error boundaries, loading overlays, pending actions). Along with this, a full demo was added to more easily demonstrate terra-application features.
* ApplicationIntlContext was created and provided from ApplicationBase to provide a future-facing Context solution for intl within the application. Trying to abstract away react-intl a little bit; won't be able to fully do so, but this should keep people from having to rip out `injectIntl` all over the place when we eventually upgrade to v3.
* Suspense components were added throughout to help future code-splitting efforts.


mjhenkes (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -106,19 +107,23 @@ const ApplicationBase = ({
         translationsLoadingPlaceholder={translationsLoadingPlaceholder}
         locale={locale}
       >
-        <ActiveBreakpointProvider>
-          <ApplicationErrorBoundary>
+        <ApplicationIntlProvider>
```
emilyrohrbough: Terra-base will render terra-i18n which will render the terra-i18n provider so we shoudn't need to define a new one

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * The Object (or function that returns an Object) that specifies the messages
+   * used to prompt the user when navigation items are selected while NavigationPrompts
+   * are rendered by the ApplicationNavigation content.
+   */
+  navigationPromptResolutionOptions: navigationPromptResolutionOptionsShape,
+};
+
+const ApplicationNavigation = ({
+  children,
```
emilyrohrbough, line 122: nit: there are a lot of props for this component. can we try to order in ABC order to quickly find the prop defs?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+   */
+  onSelectExtensionItem: PropTypes.func,
+  /**
+   * A configuration object with information pertaining to the application's user.
+   */
+  userConfig: userConfigPropType,
+  /**
+   * An element to render within the ApplicationNavigation menu, shifted to the drawer at the `medium` and below.
+   */
+  hero: PropTypes.element,
```
emilyrohrbough, line 45: What type of element should be supplied to hero? Should this be an image or a control of some sort or an external link?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Key/Value pairs associating a string key entry to a numerical notification count.
+   */
+  notifications: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
+  /**
+   * By default, the ApplicationNavigation component will resolve any rendered NavigationPrompts prior to
+   * communicating navigation item selections with `onSelectNavigationItem`. If `disablePromptsForNavigationItems`
+   * is provided, no NavigationPrompts are resolved when navigation items are selected.
+   */
+  disablePromptsForNavigationItems: PropTypes.bool,
```
emilyrohrbough: When would we want this to be enabled? By the testing framework?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+    }).catch((e) => { if (e) throw e; });
+  }, [disablePromptsForLogout, navigationPromptResolutionOptions, propOnSelectLogout]);
+
+  return (
+    <TerraApplicationNavigation
+      hero={hero}
+      notifications={notifications}
+      titleConfig={titleConfig}
+      navigationItems={navigationItems}
+      onSelectNavigationItem={propOnSelectNavigationItem && onSelectNavigationItem}
```
emilyrohrbough, line 173: Should this be passing the component defined `onSelectLogout`

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+      activeNavigationItemKey={activeNavigationItemKey}
```
emilyrohrbough, line 174: Do consumer have the ability to by-pass the navigation prompting by supplying a new `activeNavigationItemKey`?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+   */
+  onSelectUtilityItem: PropTypes.func,
+  /**
+   * A collection of child elements to render within the ApplicationNavigation body.
+   */
+  children: PropTypes.node,
+  /**
+   * Key/Value pairs associating a string key entry to a numerical notification count.
+   */
+  notifications: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
```
emilyrohrbough: What values should these be passes as? Should this be private or is there more docs that we can link in the description?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,192 @@
...
+      extensionItems={extensionItems}
+      onSelectExtensionItem={onSelectExtensionItem}
+      utilityItems={utilityItems}
+      onSelectUtilityItem={onSelectUtilityItem}
+      onSelectSettings={onSelectSettings}
+      onSelectHelp={onSelectHelp}
+      onSelectLogout={propOnSelectLogout && onSelectLogout}
+      onDrawerMenuStateChange={onDrawerMenuStateChange}
+    >
+      <ApplicationLoadingOverlayProvider>
```
emilyrohrbough, line 185: does this overlay cover the navigation items as well?

src/application-navigation/ApplicationNavigation.jsx
tbiethman: They do, and we can't really prevent them from doing so. They would have to render their own NavigationPromptCheckpoint around ApplicationNavigation if there was an alternative way to change the selected navigation items.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: I'm not sure I follow. Logout is passed below. These boolean operators ensure that the internal functions are only provided to TerraApplicationNavigation if the associated prop is defined; we'd be seeing them all the time otherwise.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: Yeah that's fair. How do we feel about the documented prop table order? I've always tended to order things in either logical groups or a rough order of "importance" (required or not).

src/application-navigation/ApplicationNavigation.jsx
tbiethman: It covers everything below the header; this is documented here: https://github.com/cerner/terra-application/pull/5/files#diff-487c8d381b9f7f279bd497d2373b0701R34

My big concern here was allowing the user access to Logout and other controls while a component is loading. It means that the user could theoretically navigate away from the loading component, but consumers could pair the overlay with a NavigationPrompt to help prevent that if necessary.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: By default, ApplicationNavigation assumes that changing primary navigation items is always a destructive action for the page contents. This is how we expect people to consume the component, and this is why we have the prompts enabled by default.

However, depending on how an application is built, changing navigation items may not actually result in data loss. This prop exists to allow consumers the option of disabling the default prompts and handling the prompts themselves.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: I can flesh out the comment some more for this since I've pared down the examples quite a bit that exist in the terra-application-navigation package. 

But like the comment says here, this is just an object that has keys that correspond to navigation item/extension/utility item keys associated to a number; the number will display alongside the appropriate element in a little notification bubble.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: Good question. This is one of those things that is still a little hazy so we left the API open. From the examples I've seen, it's typically an image of some kind, but we don't know enough to make things too strict here.

src/application-base/ApplicationBase.jsx
tbiethman: The ApplicationIntlProvider here (and the ApplicationIntlContext.Provider that it renders) are different than the IntlProvider provided by terra-base. The ApplicationIntlProvider forwards the intl value from the IntlProvider through a modern Context. 

With Terra as a whole being stuck on react-intl v2.9 for the foreseeable future, I didn't care for the idea of consumers forced to interface with the legacy context APIs just for intl. With this in place, consumer's get the benefit of a cleaner API that's more consistent with all the other context-based APIs that terra-application introduces. Consumers of this should also feel no effects from the eventual bump to react-intl v3 (and the deprecation of `injectIntl`).

The new ApplicationIntlContext sits adjacent to the existing react-intl implementation. All the existing Terra components work the same as they do today. Consumers could opt to use react-intl directly, and it'd work just fine. But for a component that already has a dependency on terra-application, they'd get everything they need for i18n for free, which is nice.

I do understand that it's a divergence from the non-application Terra components today. I feel like the benefits outweigh the drawbacks in that regard, but I'm open to input here.

src/application-navigation/ApplicationNavigation.jsx
emilyrohrbough: Can we make a note of it in the description or docs that they will be by-passing the navigation promptings?

src/application-navigation/ApplicationNavigation.jsx
emilyrohrbough: Oh I see. This isn't passing  a boolean, it's only passing the function if propOnSelectNavigationItem is defined

src/application-navigation/ApplicationNavigation.jsx
emilyrohrbough: Thanks for the link!

src/application-base/ApplicationBase.jsx
emilyrohrbough: Is this something we should be providing from terra-i18n instead of from terra-application?

Could this caused the situation were components developed and tested in terra-dev-site has access to intl via context, but then the deployed standalone components don't actually have access to it since they aren't wrapped by terra-application?

tbiethman: I have thought about that. Doing so would introduce peer dependencies on terra-i18n for all packages that need translations, which wouldn't be passive in the short term. We could also include it directly within terra-base, but that would re-introduce dependencies on terra-base where we had them previously.

I think we could still make a decision to do so in the future, but even so, I'd expect terra-application to provide that new intl context, even if it's just a passthrough at that point. The API from terra-application can remain constant and passive for consumers of terra-application.

The alternative to what I have here would be to do nothing in the short term. I'd prefer to get folks on the right path and not set them up with technical debt from the get go.

emilyrohrbough: If terra-application is providing the context, are we assuming terra-base will be deprecated or only provided for base styles?

tbiethman: > Could this caused the situation were components developed and tested in terra-dev-site has access to intl via context, but then the deployed standalone components don't actually have access to it since they aren't wrapped by terra-application?

Components that don't have a peerDependency on terra-application wouldn't use this context, which would be Terra's vast majority of components. They'd still be using react-intl directly (unless we decide to do something different in the future like mentioned above). We're still relying on there being a (single) react-intl instance in the bundle.

tbiethman: Sorry these comments are out of order, I should refresh more often.

> If terra-application is providing the context, are we assuming terra-base will be deprecated or only provided for base styles?

I don't want to make any assumptions about terra-base. I think terra-base continues to have value as the barest-bones provider of Terra's global requirements (global styles and i18n) for all components in its ecosystem. 

Applications that want to only use Terra's non-framework components (say they really like Button or something) shouldn't be required to use terra-application's ApplicationBase and its associated complexities. In these cases, having terra-base available as a stripped down option seems valuable.

emilyrohrbough: Talked offline about this. This is the route we want to go! 👍

tbiethman: :+1: I am going to revise the docs around the ApplicationIntlContext a bit to make consumption expectations more clear.

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+  onDrawerMenuStateChange: PropTypes.func,
+  /**
+   * A function to be executed upon the selection of an extensions item.
+   * Ex: `onSelectExtensionsItem(String selectedUtilityItemKey, Object metaData)`
+   */
+  onSelectExtensionItem: PropTypes.func,
+  /**
+   * A function to be executed upon the selection of the Help utility item.
+   * If `onSelectHelp` is not provided, the Help utility item will not be rendered.
+   * Ex: `onSelectLogout()`
```
ryanthemanuel: Should this ex be onSelectHelp()?

ryanthemanuel (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -106,19 +107,23 @@ const ApplicationBase = ({
...
+                    {children}
+                  </Suspense>
+                </ApplicationErrorBoundary>
               </ApplicationLoadingOverlayProvider>
             </NavigationPromptCheckpoint>
-          </ApplicationErrorBoundary>
-        </ActiveBreakpointProvider>
+          </ActiveBreakpointProvider>
+        </ApplicationIntlProvider>
       </Base>
```
emilyrohrbough, line 127: Any changes we might want a top-level error boundary in case an error occurs before loading their content?

src/application-intl/index.js
```diff
@@ -0,0 +1,4 @@
+import ApplicationIntlContext from './ApplicationIntlContext';
+import ApplicationIntlProvider from './ApplicationIntlProvider';
+
+export { ApplicationIntlContext, ApplicationIntlProvider };
```
emilyrohrbough: Should we be exporting the context Shape from the `ApplicationIntlContext ` as well?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+   * A string key representing the currently active navigation item. This value should match one of the item keys provided in the
+   * `navigationItems` array.
+   */
+  activeNavigationItemKey: PropTypes.string,
+  /**
+   * A collection of child elements to render within the ApplicationNavigation body.
+   */
+  children: PropTypes.node,
+  /**
+   * By default, the ApplicationNavigation component will resolve any rendered NavigationPrompts prior to
```
emilyrohrbough: resolve any rendered NavigationPrompts -> resolve any **registered** NavigationPrompts

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+   * communicating navigation item selections with `onSelectNavigationItem`. If `disablePromptsForNavigationItems`
+   * is provided, no NavigationPrompts are resolved when navigation items are selected.
+   */
+  disablePromptsForNavigationItems: PropTypes.bool,
+  /**
+   * A configuration object with information specifying the creation of the Extension buttons rendered within the
+   * ApplicationNavigation header.
+   */
+  extensionItems: extensionItemsPropType,
```
emilyrohrbough, line 41: Are these expected to be buttons or any react content?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * An element to render within the ApplicationNavigation menu, shifted to the drawer at the `medium` and below.
```
emilyrohrbough: to the drawer at the `medium` **breakpoint** and below.

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+   */
+  hero: PropTypes.element,
+  /**
+   * An array of configuration objects with information specifying the creation of navigation items. These items
+   * are rendered within the ApplicationNavigation header at larger breakpoints and within the drawer menu at smaller breakpoints.
```
emilyrohrbough, line 48: > larger breakpoints and within the drawer menu at smaller breakpoints.

the `hero` prop description listed a specific breakpoint. Are those these same here?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+  navigationItems: navigationItemsPropType,
+  /**
+   * The Object (or function that returns an Object) that specifies the messages
+   * used to prompt the user when navigation items are selected while NavigationPrompts
+   * are rendered by the ApplicationNavigation content.
+   */
+  navigationPromptResolutionOptions: navigationPromptResolutionOptionsShape,
+  /**
+   * Key/Value pairs associating a string key entry to a Number notification count. The keys must correspond to a
+   * navigationItem or extensionItem key provided through their associated props.
```
emilyrohrbough, line 59: Is the expected prop
key - the value of the navigationItem/extensionItem key that wants to set a notification
value - notification count
?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+   */
+  notifications: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
+  /**
+   * Callback function triggered on Drawer Menu state change
```
emilyrohrbough: missing `.` (you've been pretty consistent 😄 )

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * A function to be executed upon the selection of the Help utility item.
+   * If `onSelectHelp` is not provided, the Help utility item will not be rendered.
+   * Ex: `onSelectLogout()`
+   */
+  onSelectHelp: PropTypes.func,
+  /**
+   * A function to be executed upon the selection of the Logout action button.
+   * If `onSelectLogout` is not provided, the Logout action button will not be rendered.
+   * Ex: `onSelectLogout()`
```
emilyrohrbough, line 83: It might be helpful to mention this will be called once all NavigationPrompts have beee resolved. it was talked about above though so maybe not?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+   * A function to be executed upon the selection of a navigation item.
+   * Ex: `onSelectNavigationItem(String selectedNavigationItemKey, Object metaData)`
+   */
+  onSelectNavigationItem: PropTypes.func,
+  /**
+   * A function to be executed upon the selection of the Settings utility item.
+   * If `onSelectSettings` is not provided, the Settings utility item will not be rendered.
+   * Ex: `onSelectSettings()`
+   */
+  onSelectSettings: PropTypes.func,
```
emilyrohrbough, line 98: What type of settings could be in here? For example, if there is local switching, this would cause a full re-rendering of the content lower in the stack and likely blow away the current state of the content.

Would we expect navigation prompting or the application to handle this type of updating?

emilyrohrbough: Actually thinking about this, this would only update the locale strings, not necessarily lose the current state.

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * A function to be executed upon the selection of a custom utility item.
+   * Ex: `onSelectUtilityItem(String selectedUtilityItemKey, Object metaData)`
+   */
+  onSelectUtilityItem: PropTypes.func,
```
emilyrohrbough, line 104: I'm noticing some of these callback are shortened to `onSelectHelp` or `onSelectionSettings`, where some are `onSelectNavigationItem` or `onSelectUtilityItem`.  Would it make sense to match these?

src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const onSelectNavigationItem = useCallback((selectedItemKey) => {
+    if (disablePromptsForNavigationItems) {
+      propOnSelectNavigationItem(selectedItemKey);
+      return;
+    }
+
+    navigationPromptCheckpointRef.current.resolvePrompts(navigationPromptResolutionOptions).then(() => {
+      propOnSelectNavigationItem(selectedItemKey);
+    }).catch((e) => { if (e) throw e; });
```
emilyrohrbough, line 153: Why catch if we are just going to re-throw it?

src/application-intl/ApplicationIntlContext.jsx
```diff
@@ -0,0 +1,9 @@
+import { createContext } from 'react';
+import { intlShape } from 'react-intl';
+
+const ApplicationIntlContext = createContext();
+
+const contextShape = intlShape;
+
+export default ApplicationIntlContext;
+export { contextShape };
```
emilyrohrbough, line 9: I'm surprised the linter isn't complaining about it needing a new line at the end of the file

src/application-navigation/index.js
```diff
@@ -0,0 +1,10 @@
+import {
+  titleConfigPropType, navigationItemsPropType, extensionItemsPropType, utilityItemsPropType, userConfigPropType,
+} from 'terra-application-navigation/lib/utils/propTypes';
+import ApplicationNavigation from './ApplicationNavigation';
+
+export default ApplicationNavigation;
+export {
+  titleConfigPropType, navigationItemsPropType, extensionItemsPropType, utilityItemsPropType, userConfigPropType,
```
emilyrohrbough, line 8: why export the `terra-application-navigation` prop types instead of the proptypes for `./ApplicationNavigation`?

tests/jest/application-error-boundary/index.test.js
```diff
@@ -0,0 +1,7 @@
+import ApplicationErrorBoundary from '../../../src/application-error-boundary';
+
+describe('application-error-boundary/index', () => {
```
emilyrohrbough, line 3: I think these are nice sanity checks of exports, but if you look at the generated jest coverage, I don't think these tests affect it....

tests/jest/application-navigation/ApplicationNavigation.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+        onDrawerMenuStateChange={jest.fn()}
+      >
+        <div>test child content</div>
+      </ApplicationNavigation>,
+    );
+
+    expect(shallowComponent).toMatchSnapshot();
+  });
+
+  it('should render with all props', () => {
```
emilyrohrbough: looks like this is a duplicate of [the test before this](https://github.com/cerner/terra-application/pull/5/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.jsx&file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.snap#diff-611946ada81ee56923728d2306a6ee54R14)

tests/jest/application-navigation/ApplicationNavigation.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+        }}
+        navigationItems={[{ text: 'test-text', key: 'my-test-key' }]}
+        extensionItems={[{ text: 'test-text-1', key: 'my-test-key-1', icon: <span>my icon</span> }]}
+        utilityItems={[{ text: 'test-text-2', key: 'my-test-key-2', icon: <span>my icon</span> }]}
+        notifications={{ 'my-test-key-1': 3 }}
+        navigationRenderFunction={jest.fn()}
+        onSelectNavigationItem={jest.fn()}
+        onSelectExtensionItem={jest.fn()}
+        onSelectUtilityItem={jest.fn()}
+        onSelectSettings={jest.fn()}
```
emilyrohrbough, line 37: Did we want to ensure the navigation prompting was being handled and/or bypassed as expected since we wrap the `onSelectNavigationItem` and `onSelectLogout` functions that consumers will be providing?

tests/wdio/application-navigation/application-navigation-spec.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    it('prompts user when selecting logout', () => {
+      browser.click('#pending-action-toggle');
+      browser.click('[data-application-header-utility="true"]');
+      browser.waitForVisible('[data-terra-popup-content="true"]');
+      browser.click('[data-terra-popup-content="true"] button');
+
+      browser.waitForVisible('[data-terra-notification-dialog="true"]');
+      Terra.validates.element('4. primary navigation item prompt', { selector });
```
emilyrohrbough:
```suggestion
      Terra.validates.element('4. logout prompt', { selector });
```

src/terra-dev-site/doc/application/how-to.2/manage-pending-actions.doc.mdx
```diff
@@ -27,9 +27,11 @@ const ExampleComponent = () => {
...
+Within `terra-application`, the following components respond to the presence of NavigationPrompts:
 
-[ModalManager](/components/terra-application/application/reference/components/modal-manager) and [SlidePanelManager](/components/terra-application/application/reference/components/slide-panel-manager) will check for the presence of NavigationPrompts within their disclosed components and prompt the user if necessary before allowing the components to be dismissed. Please see their respective reference pages for more information.
+* [ApplicationBase](/components/terra-application/application/reference/components/application-base) - NavigationPrompts rendered within children will result in prompting prior to window unload events.
+* [ModalManager](/components/terra-application/application/reference/components/modal-manager)/[SlidePanelManager](/components/terra-application/application/reference/components/slide-panel-manager) - NavigationPrompts rendered within disclosed components will result in prompting prior to disclosure dismissal.
```
emilyrohrbough: **If** NavigationPrompts **are** rendered within disclosed components, **the user** will ~result in prompting~ **be prompted before the content can be dismissed** ~prior to disclosure dismissal.~

src/terra-dev-site/doc/application/how-to.2/manage-pending-actions.doc.mdx
```diff
@@ -27,9 +27,11 @@ const ExampleComponent = () => {
...
+* [ApplicationNavigation](/components/terra-application/application/reference/components/application-navigation) - NavigationPrompts rendered within children will result in prompting prior to executing callbacks for primary navigation item and 'Logout' selection.
```
emilyrohrbough: The user will be promoted to accept or reject primary navigation and logout selection if NavigationPrompts have been rendered within the ApplicationNavigation's children.

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
```diff
@@ -1,25 +1,26 @@
 # How To Use Translated Strings
 
-Translated strings can be retrieved using the `intl` object provided by `injectIntl` HOC-generator provided by [react-intl](https://github.com/formatjs/react-intl/blob/master/docs/API.md#injectintl).
+Translated strings can be retrieved using the [ApplicationIntlContext](/components/terra-application/application/reference/contexts/application-intl-context). An `ApplicationIntlContext.Provider` is configured and provided by [ApplicationBase](/components/terra-application/application/reference/components/application-base). The strings will be translated based upon the `locale` prop provided to ApplicationBase.
```
emilyrohrbough: The **pre-aggregated translation** strings will be ~translated~ **loaded** based upon the `locale` prop provided to ApplicationBase.


It'd be worth mention terra-i18n's loading strategy and the fallback strategy to english, even if it's linking to the docs.

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
```diff
@@ -1,25 +1,26 @@
...
+
+In addition to the ApplicationIntlContext, the components provided by react-intl can be used directly. However, the ApplicationIntlContext can be used by Hooks-based components to access i18n capabilities without having to use the `injectIntl` HOC-generator from react-intl.
```
emilyrohrbough: `ApplicationIntlContext`, the components provided by `react-int`

emilyrohrbough: Are we trying to pushing this stategy? If so, we should switch the order of these sentances:

The ApplicationIntlContext **is useful** because it can be used by Hooks-based components to access i18n capabilities without having to use the `injectIntl` HOC-generator from react-intl. However, components  provided by react-intl can be used directly.

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
```diff
@@ -1,25 +1,26 @@
...
 
 ```jsx
-import React from 'react';
-import { injectIntl, intlShape } from 'react-intl';
+import React, { useContext } from 'react';
+import { ApplicationIntlContext } from 'terra-application/lib/application-intl';
+import { FormattedMessage } from 'react-intl';
+
+const ExampleComponent = () => {
+  const applicationIntl = useContext(ApplicationIntlContext);
```
emilyrohrbough, line 18: I'm looking at the react-intl v3 upgrade guide. Their example for utilizing intl with hooks is actually a custom hook called `useIntl`. What if we followed this strategy to make the v3 upgrade feel seem-less when we swap ours out for theirs?

https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#new-useintl-hook-as-an-alternative-of-injectintl-hoc

[src code reference](https://github.com/formatjs/react-intl/blob/master/src/components/useIntl.ts)

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
```diff
@@ -1,25 +1,26 @@
 # How To Use Translated Strings
 
-Translated strings can be retrieved using the `intl` object provided by `injectIntl` HOC-generator provided by [react-intl](https://github.com/formatjs/react-intl/blob/master/docs/API.md#injectintl).
+Translated strings can be retrieved using the [ApplicationIntlContext](/components/terra-application/application/reference/contexts/application-intl-context). An `ApplicationIntlContext.Provider` is configured and provided by [ApplicationBase](/components/terra-application/application/reference/components/application-base). The strings will be translated based upon the `locale` prop provided to ApplicationBase.
+
+In addition to the ApplicationIntlContext, the components provided by react-intl can be used directly. However, the ApplicationIntlContext can be used by Hooks-based components to access i18n capabilities without having to use the `injectIntl` HOC-generator from react-intl.
 
 ```jsx
```
emilyrohrbough, line 13: I think this example is useful for the how-to, however, looking at this git-diff, I think it'd be helpful to pull this diff forward for consumers that want to upgrade existing content to use this new API and want to know how that should be done. We could render thee example in a collapsed section labeled something like "Upgrade to use ApplicationIntlContext"

src/terra-dev-site/doc/application/reference.1/components/ApplicationBase.doc.mdx
```diff
@@ -20,7 +20,7 @@ import ApplicationBase from 'terra-application/lib/application-base';
 
 ### I18n
 
-ApplicationBase exposes a prop from defining the application's locale. ApplicationBase renders an IntlProvider (from `react-intl`) around its children. Children can access intl APIs through the legacy Context API provided by `react-intl`.
+ApplicationBase exposes a prop for defining the application's locale. ApplicationBase renders an [IntlProvider](https://engineering.cerner.com/terra-ui/components/terra-i-18-n/i-18-n/i-18-n) (from [react-intl](https://github.com/formatjs/react-intl)) as well as a [ApplicationIntlContext](/components/terra-application/application/reference/contexts/application-intl-context).Provider around its children. Children can access the intl APIs by using the [ApplicationIntlContext](/components/terra-application/application/reference/contexts/application-intl-context)) or by using react-intl's provided components directly.
```
emilyrohrbough: I think it would be helpful to expand the description that the Intl Provider manages switching the local and loading new local data for the application. Just some context to why Base is providing them and what they are.

emilyrohrbough: > or by using react-intl's provided components directly.
 
probably feels redundant, but can we link to react-intl's wiki page here for quick ref since it's the first mention on this doc

src/terra-dev-site/doc/application/reference.1/components/ApplicationBase.doc.mdx
```diff
@@ -40,4 +40,8 @@ ApplicationBase renders an [ApplicationLoadingOverlayProvider](/components/terra
...
+ApplicationBase monitors its children for the presence of rendered [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt). ApplicationBase will ensure that the user is prompted using a browser-native confirmation dialog prior to any window unload event if any [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt) are rendered at the time of the unload request.
+
+### Code Splitting
+
+ApplicationNavigation renders a [Suspense](https://reactjs.org/docs/code-splitting.html#reactlazy) component around its children. If any child components or their descendants are lazy loaded using [React.lazy](https://reactjs.org/docs/code-splitting.html#reactlazy), a sensible fallback will be rendered while the component is being retrieved. The fallback will render below the ApplicationNavigation header and allow the header to remain interactable.
```
emilyrohrbough: > any child components or their descendants

Really this will render a loading indicator on the full tree if React.Lazy is used at a way lower level?

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
+import PropsTable from 'terra-application/lib/application-navigation/ApplicationNavigation?dev-site-props-table';
+import ApplicationNavigationExample from 'terra-application/lib/terra-dev-site/doc/application/reference.1/components/example/ApplicationNavigationExample?dev-site-example';
+
+# ApplicationNavigation
+
+The ApplicationNavigation component provides a responsive layout and a variety of controls for application navigation.
```
emilyrohrbough: 😄 can we go into detail how it's responsive and what types of controls are provided? I know the props table will give options, but it'd be easier to understand what you get from this if you can see a face-up list

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+<PropsTable />
+
+## Example
+
+<ApplicationNavigationExample />
+
+## Features
+
+### Exception Handling
+
```
emilyrohrbough, line 249: ~ApplicationBase~ **ApplicationNavigation** will catch the exceptions and

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+ApplicationNavigation renders an [ApplicationErrorBoundary](/components/terra-application/application/reference/components/application-error-boundary) around its children. If the children throw exceptions during React lifecycle functions, ApplicationBase will catch the exceptions and render a styled error component in place of the children. 
+
+This error component will render below ApplicationNavigation's header, leaving the user with full access to all navigation controls after an error occurs.
```
emilyrohrbough, line 253: This makes me think, can the settings controls be used to update the ApplicationNavigation's children content? If so, are these disabled when an error is thrown or how should these be managed by the implementer?

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+### Loading Overlays
+
+ApplicationNavigation renders an [ApplicationLoadingOverlayProvider](/components/terra-application/application/reference/components/application-loading-overlay) around its children. Any [ApplicationLoadingOverlay](/components/terra-application/application/reference/components/application-loading-overlay) components rendered by children will result in loading overlays being rendered by ApplicationNavigation.
+
+The loading overlay will render below ApplicationNavigation's header, leaving the user with full access to all navigation controls while the loading overlay is presented.
+
+### Pending Actions
+
+ApplicationNavigation monitors its children for the presence of rendered [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt). ApplicationNavigation will ensure that the user is prompted prior to executing callbacks for primary navigation item and 'Logout' utility selection if any [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt) are rendered at the time of item selection.
```
emilyrohrbough: ApplicationNavigation will ensure that the user is prompted ~prior to executing callbacks for~ **when a** primary navigation ~item and~ **or 'Logout' ~utility~ selection **is made** if any [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt) ~are~ **have been** rendered at the time of ~item~ **the** selection.

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+If the children within ApplicationNavigation can be changed by a mechanism other than navigation item selection, a NavigationPromptCheckpoint should be used directly to check for pending actions prior to the pending child's removal.
```
emilyrohrbough: We should link docs to NavigationPromptCheckpoint

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
emilyrohrbough: ApplicationNavigation monitors its children for the presence of rendered [NavigationPrompts](/components/terra-application/application/reference/components/navigation-prompt) **by wrapping a NavigationPromptCheckpoint around the children**

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+> Note: This default prompting can be disabled with the `disablePromptsForNavigationItems` and `disablePromptsForLogout` props if a custom implementation is required.
+
+### Code Splitting
+
+ApplicationNavigation renders a [Suspense](https://reactjs.org/docs/code-splitting.html#reactlazy) component around its children. If any child components or their descendants are lazy loaded using [React.lazy](https://reactjs.org/docs/code-splitting.html#reactlazy), a sensible fallback will be rendered while the component is being retrieved. The fallback will render below the ApplicationNavigation header and allow the header to remain interactable.
```
emilyrohrbough: > a sensible fallback 

Isn't this the ApplicationLoadingOverlay?

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
+import PropsTable from 'terra-application/lib/application-navigation/ApplicationNavigation?dev-site-props-table';
+import ApplicationNavigationExample from 'terra-application/lib/terra-dev-site/doc/application/reference.1/components/example/ApplicationNavigationExample?dev-site-example';
+
+# ApplicationNavigation
+
+The ApplicationNavigation component provides a responsive layout and a variety of controls for application navigation.
+
+## Usage
```
emilyrohrbough, line 8: we should drop the usage section since we have the code example they can reference

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Props
+
+<PropsTable />
+
+## Example
+
+<ApplicationNavigationExample />
+
+## Features
```
emilyrohrbough, line 246: Looking at the deployed site, I think it'd be nice to read the features before the prop types (same comment for all docs)

src/terra-dev-site/doc/application/reference.1/contexts/ApplicationIntlContext.doc.mdx
```diff
@@ -0,0 +1,13 @@
+# ApplicationIntlContext
+
+The ApplicationIntContext defines an interface for access to the framework's i18n APIs.
+
+## Context Value
+
+The value of the ApplicationIntlContext is an object containing i18n APIs. Please review the IntlShape as defined by [react-intl](https://github.com/formatjs/react-intl/blob/master/docs/API.md#intlshape).
```
emilyrohrbough: Can we add an example on how we'd use this?

Edit: can we direct to the use-translated-strings how-to guide for usage?

emilyrohrbough (commented):
This could be done in a different PR, but it'd be nice to tweak/flatten this documentation a bit. Looking at the deployment, it'd be nice if we changed the `Components` tab to `Docs` and then flatten the navigation structure to something like:

Screen Shot 2019-12-07 at 4 50 32 PM

emilyrohrbough:
Demo looks good. Noticing a few small things:
1 ) clicking logout does nothing. In the wdio tests you rendered a logout page. Can we do the same?
2) you're missing navigation prompt resolution options int he modal rendered by the utility item:
Screen Shot 2019-12-07 at 4 56 25 PM

emilyrohrbough:
Of all the comments I just left, the ones with 👀  are the ones I'd like to see resolved before approving. The remaining are doc tweaks or changes we can do later. These should be these ones I care about:
- https://github.com/cerner/terra-application/pull/5#discussion_r355144454
- https://github.com/cerner/terra-application/pull/5#discussion_r355145338
- https://github.com/cerner/terra-application/pull/5#discussion_r355145487
- https://github.com/cerner/terra-application/pull/5#discussion_r355145646
- https://github.com/cerner/terra-application/pull/5#discussion_r355145831
- (bigger question) https://github.com/cerner/terra-application/pull/5#discussion_r355146789
- https://github.com/cerner/terra-application/pull/5#discussion_r355147329
- https://github.com/cerner/terra-application/pull/5#issuecomment-562894181

src/application-base/ApplicationBase.jsx
tbiethman: My thinking was that we'd want to keep as much mounted as possible in the case of a child error. I guess we could move it up immediately under Base?

src/application-intl/ApplicationIntlContext.jsx
tbiethman: They're present in this file and throughout, just not rendered here.

src/application-intl/index.js
tbiethman: Yes, good call

src/application-navigation/ApplicationNavigation.jsx
tbiethman: To follow through on your point, though. If an exposed settings option were to be destructive to ApplicationNavigation or its children, the consumer of ApplicationNavigation would be responsible for prompting the user.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: All these props are carried forward from the terra-application-navigation package. I'd prefer not to update the existing props unless there's something egregiously wrong with them.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: I didn't write this one :)

src/application-navigation/ApplicationNavigation.jsx
tbiethman: It's described in the other documentation more clearly. I'll leave it out of the props table to keep these descriptions more concise.

src/application-navigation/ApplicationNavigation.jsx
tbiethman: The NavigationPromptCheckpoint rejects its returned Promise when the user opts to reject the prompt, but it doesn't return an error with the rejection. This catch will swallow that error-free rejection and but re-throw any present errors that originated from something else.

Here I'm just trying to keep the 'Uncaught error in Promise' logs from showing up in the console when users reject the prompt.

src/application-navigation/index.js
tbiethman: The different propType shapes aren't exported from ./ApplicationNavigation. We import them there, we use them there, but we're not exporting them from there because this index file is a more likely consumption point.

src/terra-dev-site/doc/application/reference.1/components/ApplicationBase.doc.mdx
tbiethman: It would indeed. The alternative would be nothing rendering at all until that script is retrieved.

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
tbiethman: I'll go ahead and copy forward a lot of the docs from the original package. It'll give helpful context to a lot of these concepts.

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
tbiethman: The 'Usage' section for the components here really just show how to import the components. I think that can still be helpful face-up for people vs. viewing the example code.

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
tbiethman: I waffled a bit on this. I originally had things ordered the same way, but anyone using the doc site as a prop reference would be scrolling around a lot. Figured for the reference pages, the props might be more useful?

src/terra-dev-site/doc/application/reference.1/components/ApplicationNavigation.doc.mdx
tbiethman: I touched on this a bit above: https://github.com/cerner/terra-application/pull/5/files#r355158502

Settings from ApplicationNavigation should application-level, not content-level. Changing settings shouldn't be blocked by a component runtime error caught by the boundary.

tests/jest/application-navigation/ApplicationNavigation.test.jsx
tbiethman: Those workflows are covered by the webdriver tests.

tbiethman: We could test them here, but they'd be...uglier.

tests/jest/application-error-boundary/index.test.js
tbiethman: Well the coverage config was jacked in general. Fixed that, but you're right that the index files aren't tracked for coverage at all. But they are running, and they validate the exposed APIs from each directory.

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
tbiethman: We could expose an additional `useIntl` hook, but we'd still need to create our own Context to do so.

The kicker here is that I wasn't planning on creating custom hooks for the various other contexts being exposed. Do you guys find them (in general) useful? It might be awkward to offer only one. They might be easier for  consumers to test with?

src/terra-dev-site/doc/application/reference.1/components/ApplicationBase.doc.mdx
tbiethman: I rewrote this a bit to include your points.

tests/jest/application-navigation/ApplicationNavigation.test.jsx
emilyrohrbough: Yeah I'd leave it as it. I didn't see the wdio tests till later on

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
emilyrohrbough: IMO I think a custom hook is a nicity that doesn't really give much extra to consumers. I don't think we need to provided a custom hook for our context APIs. My only thought here would be that the upgrade path would be simple

src/terra-dev-site/doc/application/how-to.2/manage-pending-actions.doc.mdx
tbiethman: I've updated these to have a similar structure

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
tbiethman: I don't think we're trying to "push" it per se. It's more along the lines of "We use react-intl. You should use react-intl. If for some reason you want/need to use a Context before the react-intl upgrade, you can use the ApplicationIntlContext."

I'm going to refactor this how-to quite a bit to primarily point to the existing I18n docs on terra-ui.com. It's all useful stuff, but I don't want to dupe it here. But I'll make explicit mention here of the ApplicationIntlContext and scenarios where it might be useful.

tbiethman:
> This could be done in a different PR, but it'd be nice to tweak/flatten this documentation a bit. Looking at the deployment, it'd be nice if we changed the Components tab to Docs and then flatten the navigation structure to something like:

The root 'Application' folder is present to allow these docs to be rendered within terra-ui. The other folders I can play around with.

tbiethman:
> Demo looks good. Noticing a few small things:
> 1 ) clicking logout does nothing. In the wdio tests you rendered a logout page. Can we do the same?
> 2) you're missing navigation prompt resolution options int he modal rendered by the utility item:

1) Yep, got that in there.
2) Good catch. Now that I have the translated strings, I went ahead and set default values for the navigation prompts within the ModalManager/SlidePanelManager. I still allow overrides but the defaults are nice. So that should be fixed too.

src/terra-dev-site/doc/application/how-to.2/use-translated-strings.doc.mdx
tbiethman: I'd prefer to hold off on providing it for the time being and add it later if we need to based on consumer sentiment. Sound good to you?

emilyrohrbough (approved):
Nice job on this Tyler!

PR #13, opened 2020-02-20
Add preferred browser language fallback strategy 

### LINKED ISSUE #11: [Spike] Change locale to be optional and default to the browser locale
# Summary

Spike on changing the [locale prop](https://github.com/cerner/terra-application/blob/c64a644e453a707aa57d42cd41b59e975f35b4d3/src/application-base/ApplicationBase.jsx#L25) in application base to be optional. If a locale is provided it will be used. If a locale is not provided the locale should be interpreted from the browser. 

If no locale is provided and no locale can be interpreted from the browser the locale should follow the fallback pattern and use 'en'.

The browser locale can be discovered doing something similar to this: 

```jsx
const locale = (navigator.languages && navigator.languages[0])
               || navigator.language
               || navigator.userLanguage
               || 'en';
```

Resources: 

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

## Locales

To ensure the fallback locale obtained from the browser uses a supported locale it will need to be compared against a list of known supported locales.

This list can be obtained a few ways. There is a [default locale](https://github.com/cerner/terra-aggregate-translations/blob/master/config/i18nSupportedLocales.js#L1) list in terra-aggregate-translations. There is the terra18n.config list that users define.

These values could potentially be added to the webpack bundle using the [webpack define](https://webpack.js.org/plugins/define-plugin/) functionality. This functionality is used today to provide the BASE_PATH and CERNER_BUILD_TIME globals.

StephenEsser:
### Summary

This PR updates the locale to be an optional prop in ApplicationBase. If no locale is provided a fallback strategy will be used to obtain the preferred language from the browser. The preferred language must be included in the TERRA_AGGREGATED_LOCALES global. This global represents the known supported locale list for translations.

If no supported language can be resolved the fallback strategy will default the language to `en`.

TERRA_AGGREGATED_LOCALES is provided as a global in terra-toolkit >= v5.21.0. Those changes can be found [here](https://github.com/cerner/terra-toolkit/pull/383).

### Additional Details

The fallback strategy is intended to be very similar to the recommendation seen in the [terra-base documentation](https://engineering.cerner.com/terra-core/components/terra-base/base/base) with the addition of checking the locale against the supported aggregated locales first.

```
const locale = (navigator.languages && navigator.languages[0])
               || navigator.language
               || navigator.userLanguage
               || 'en';
```

Resolves #11 

@cerner/terra

[CONTRIBUTORS.md]: https://github.com/cerner/terra-application/blob/master/CONTRIBUTORS.md
[License]: https://github.com/cerner/terra-application/blob/master/License.md


src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (isSupported(navigator.browserLanguage)) {
+    return navigator.browserLanguage;
+  }
+
+  return DEFAULT_LOCALE;
+};
+
+export default getBrowserLocale;
+export { isSupported, filterLanguages };
```
emilyrohrbough: are we exporting these just for testing?

emilyrohrbough:
Is there any concerns that navigator is not implemented consistently across browsers? We have recently had Intl inconsistency in embedded IE and in windows server 2012 r2.

src/browser-locale/index.js
StephenEsser: Yes

StephenEsser:
This fallback strategy is because navigator is not implemented consistently across browsers. It will first try to obtain the preferred languages using the most common implementation of navigator and strategically fallback to other known navigator accessors for other browsers.

Do you have concerns in mind that are not accounted for here?

src/browser-locale/index.js
emilyrohrbough: I'm all for more test coverage, but this feels a little weird to export them for testing purposes only. If we flipped this to a class would we be able to test without exporting these? or not a good idea?

StephenEsser: Creating a class would have the same result. The functions would be exported with the class and become part of the public API.

src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return false;
+};
+
+/**
+ * Filters the user's preferred languages from the browser and returns the first language found that is included in the supported locale list.
+ * @returns {string|null} - A supported locale. Null if not found.
+ */
+const filterLanguages = () => {
+  if (navigator.languages && navigator.languages.length > 0) {
```
yuderekyu: not a big deal, but i believe we can simplify the second conditional to a length check
```suggestion
  if (navigator.languages && navigator.languages.length) {
```

tests/jest/browser-locale/index.test.js
```diff
@@ -0,0 +1,150 @@
...
+    });
+  });
+
+  describe('filterLanguages', () => {
+    beforeEach(() => {
+      Object.defineProperty(navigator, 'languages', { get() { return undefined; }, configurable: true });
+    });
+
+    afterEach(() => {
+      Object.defineProperty(navigator, 'languages', { get() { return undefined; }, configurable: true });
```
yuderekyu: curious - what's the reason we need to reset `navigator` with afterEach? it seems like the beforeEach would be enough?

StephenEsser: The intention is to be explicit about cleaning up any mock overrides. Without the afterEach cleanup the last test would leave the mocking active for any tests run there after in a different describe block. We could potentially change this to an afterAll.

src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+// The fallback locale that will be used if a supported locale is unable to be retrieved from the browser.
+const DEFAULT_LOCALE = 'en';
+
+/**
+ * Returns a boolean indicating if the locale is supported.
+ * @param {string} locale - The locale.
+ * @returns {bool} - A boolean indicating the locale is supported.
+ */
+const isSupported = (locale) => {
```
emilyrohrbough: Should we be ensuring the string format is correct? On your spike, it links to: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language

it says safari <10.2 returns country code as lowercase
> Note that in Safari on iOS prior to 10.2, the country code returned is lowercase: "en-us", "fr-fr" etc.

I don't think we support that version version but just curious if we need to consider it

src/browser-locale/index.js
emilyrohrbough: Looking at https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages which was linked on the issue, it doesn't appear `navigator.languages` is not supported by IE 10. How do we plan to match this functionality in IE 10? Or how is this being handled?

src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  return false;
+};
+
+/**
+ * Filters the user's preferred languages from the browser and returns the first language found that is included in the supported locale list.
+ * @returns {string|null} - A supported locale. Null if not found.
+ */
+const filterLanguages = () => {
```
emilyrohrbough: By leveraging this, does this mean we need to consider listening for the `languagechange_event` to ensure w are updating if the user updates their language settings once the application is open?

https://developer.mozilla.org/en-US/docs/Web/API/Window/languagechange_event

src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+ * @returns {string} - A supported browser locale. Falls back to en.
+ */
+const getBrowserLocale = () => {
+  const preferredLocale = filterLanguages();
+
+  if (preferredLocale) {
+    return preferredLocale;
+  }
+
+  if (isSupported(navigator.language)) {
```
emilyrohrbough: According to https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-language

`navigator.language` should be returning the plausable language or most preferred language. Should we be checking this value before filtering the languages?

src/browser-locale/index.js
```diff
@@ -0,0 +1,64 @@
...
+    return navigator.language;
+  }
+
+  if (isSupported(navigator.userLanguage)) {
+    return navigator.userLanguage;
+  }
+
+  if (isSupported(navigator.browserLanguage)) {
+    return navigator.browserLanguage;
+  }
```
emilyrohrbough: Are `navigator.userLanguage` or `navigator.browserLanguage` specific to IE?

src/browser-locale/index.js
StephenEsser: I'd say no. In the rare edge case a user changes their preferred language while using the application I would expect there to be a page refresh to pick up the new language preferences. Changing the locale from this event would re-render the entire app.

src/browser-locale/index.js
StephenEsser: This is handled by the fallback strategy in the `getBrowserLocale`. It checks each available navigator accessor until a supported locale is found. If navigator.languages is undefined or does not contain a supported locale it will fallback to `language`, `userLanguage`, `browserLanguage`, and eventually `en` if no locale has been found.

src/browser-locale/index.js
StephenEsser: I don't think so. Browsing through various articles it suggests that the navigator.language is not a reliable way to obtain the user preference. From the Mozilla documentation we want the user defined languages to be used as first priority:

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage

src/browser-locale/index.js
StephenEsser: Yes, if you open the dropdown from the Mozilla documentation there are references to using these in IE:

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

image

src/browser-locale/index.js
emilyrohrbough: I see. I missed that logic flow with how this is broken out

mjhenkes (approved):

emilyrohrbough (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -9,20 +9,24 @@ import ApplicationErrorBoundary from '../application-error-boundary';
...
+const browserLocale = getBrowserLocale();
+
 const propTypes = {
   /**
    * The components to render within ApplicationBase.
    */
   children: PropTypes.node.isRequired,
   /**
    * The locale name to be used to load translated messages.
+   * If not provided the preferred language from the browser will be attempted to be used.
```
tbiethman: Nitpicking the props table description here:
```suggestion
   * If the `locale` prop is not provided, the preferred language from the browser will be used.
```

src/browser-locale/private/getBrowserLocale.js
```diff
@@ -0,0 +1,64 @@
+/* global TERRA_AGGREGATED_LOCALES */
```
emilyrohrbough: Could we move this logic into application base?
```
./application-base
  /private
      getBrowserLocale
```
You said it shoudln't be consumed anywhere else right?

StephenEsser: That makes sense to me: https://github.com/cerner/terra-application/pull/13/commits/a9c1c89a5086c5d798f6377f507de1f3b8c9e2f0

emilyrohrbough (approved):

mjhenkes (approved):

tbiethman:
Do you want to throw a comment in the README regarding the private directories?

mjhenkes:
I added a blurb for private directories here:
https://github.com/cerner/terra-ui/pull/252/commits/ddbe91e5bc20df8a54286f14ee8e640ea372b40a

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
```diff
@@ -42,13 +42,13 @@ Terra.describeViewports('SlidePanelManager', ['large'], () => {
...
   it('does not display a navigation prompt when attempting to go back to the nested panel', () => {
-    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] .header-back-button');
+    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] [class*="back-button"]');
     expect(browser.isExisting('[class*="notification-dialog"]')).to.be.false;
     browser.waitForVisible('[class*="slide-group"] #DemoContainer-2 .disclose-tiny');
   });
 
   it('displays a navigation prompt when attempting to go back to the root panel', () => {
-    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] .header-back-button');
+    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] [class*="back-button"]');
```
emilyrohrbough: @mjhenkes  you should use the new selector added recently: https://github.com/cerner/terra-core/pull/2862/files

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
```diff
@@ -61,7 +61,7 @@ Terra.describeViewports('SlidePanelManager', ['large'], () => {
   });
 
   it('displays navigation prompt when attempting to close the panel', () => {
-    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] .header-close-button');
+    browser.click('[class*="SlidePanel"] [class*="ActionHeader"] .close-button');
```
emilyrohrbough:
```suggestion
    browser.click('[class*="SlidePanel"]  [data-terra-action-header="close-button"]');
```

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
```diff
@@ -14,7 +14,7 @@ Terra.describeViewports('SlidePanelManager', ['large'], () => {
...
```
emilyrohrbough:
```suggestion
    browser.click('[class*="SlidePanel"]  [data-terra-action-header="close-button"]');
```

tbiethman (approved):

PR #14, opened 2020-03-06
[select/filter] ProtoTypes for dialog and native options.

dkasper-was-taken:
### Summary
Options for displays single/multi selects in stand alone and filterable variants.

### Additional Details
Examples for:
Multi-select Filtered Dialog
Single-select Filtered Dialog
Multi-select Non-filtered Dialog
Single-select Non-filtered Inline
Single Select Native

The automatic scrolling is missing, and not all bounds checking is in place, but most should be functional.

Deployment:
https://terra-applic-new-select-z2fqp5.herokuapp.com/components/terra-application/application/demo

@cerner/terra
@ryanthemanuel @mjhenkes


emilyrohrbough:
on the deployment, I'm seeing this in the console when trying to select something from the demo:
```
Uncaught ReferenceError: metaData is not defined
    at g (FancyList.js:78)
    at Object.l (react-dom.production.min.js:14)
    at p (react-dom.production.min.js:14)
    at react-dom.production.min.js:14
    at y (react-dom.production.min.js:15)
    at it (react-dom.production.min.js:52)
    at ot (react-dom.production.min.js:51)
    at lt (react-dom.production.min.js:52)
    at pt (react-dom.production.min.js:56)
    at oe (react-dom.production.min.js:287)
```

PR #21, opened 2020-04-07
Providing Theme Context/Update theme provider.

mjhenkes:
### Summary
The ThemeContext defines an interface for access to the framework's current theme.

This context can be used to apply any theme related changes to a component.

The most common use would be to apply a theme class to the root tag of the component to apply theme variables. See below for an example.

As a part of updating theme provider the `themeIsGlobal` prop has been removed. While this traditionally would be considered non-passive, this prop was broken and I would consider it's presence a bug.

### Deployment Link
https://terra-applic-theme-cont-kndjvz.herokuapp.com/home

### Testing
This has been tested using jest as well as having been tested in the terra-framework package.

### Additional Details
I've also added theme switching to this repo.

@cerner/terra

### DIFF
--- a/src/application-base/ApplicationBase.jsx
+++ b/src/application-base/ApplicationBase.jsx
@@ -1,9 +1,12 @@
-import React, { useRef, useEffect, Suspense } from 'react';
+import React, {
+  useRef, useEffect, Suspense, useMemo,
+} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Base from 'terra-base';
 import ThemeProvider from 'terra-theme-provider';
 import { ActiveBreakpointProvider } from 'terra-breakpoints';
+// import ThemeContextProvider from 'terra-theme-context/lib/ThemeContextProvider';
 
 import ApplicationErrorBoundary from '../application-error-boundary';
 import ApplicationLoadingOverlay, { ApplicationLoadingOverlayProvider } from '../application-loading-overlay';
@@ -52,10 +55,6 @@ const propTypes = {
    */
   themeName: PropTypes.string,
   /**
-   * If provided, the theme styles are applied to the entire document.
-   */
-  themeIsGlobal: PropTypes.bool,
-  /**
    * By default, the elements rendered by ApplicationBase are fit to the Application's parent using 100% height.
    * If `fitToParentIsDisabled` is provided, the Application will render at its intrinsic content height and
    * overflow potentially overflow its parent.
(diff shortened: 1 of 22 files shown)


src/application-base/ApplicationBase.jsx
```diff
@@ -100,36 +99,40 @@ const ApplicationBase = ({
...
-      className={cx('application-theme-provider', { fill: !fitToParentIsDisabled })}
-      themeName={themeName}
-      isGlobalTheme={themeIsGlobal}
-    >
-      <Base
-        customMessages={customTranslatedMessages}
-        translationsLoadingPlaceholder={translationsLoadingPlaceholder}
-        locale={locale || browserLocale}
+    <div className={cx('application-base', { fill: !fitToParentIsDisabled })}>
+      <ThemeProvider
```
emilyrohrbough, line 106: Its almost too bad we updated this to be a react fragment

src/application-base/ApplicationBase.module.scss
```diff
@@ -1,15 +1,9 @@
...
-    
-  .application-theme-provider.fill {
-    height: 100%;
-  } 
+  }
 
   .application-base.fill {
     height: 100%;
-    overflow: auto;
-    -webkit-overflow-scrolling: touch;
```
emilyrohrbough, line 13: do we still need these styles?

src/application-base/ApplicationBase.jsx
```diff
@@ -51,10 +54,6 @@ const propTypes = {
    * The name of the theme to apply to the application using terra-theme-provider.
    */
   themeName: PropTypes.string,
-  /**
-   * If provided, the theme styles are applied to the entire document.
-   */
-  themeIsGlobal: PropTypes.bool,
```
emilyrohrbough, line 57: will want to add this prop removal in the changelog

emilyrohrbough (approved):
Looks good. Just need to add a changelog entry

src/application-base/ApplicationBase.module.scss
mjhenkes: Funny thing, we weren't applying them before. So I checked with @tbiethman, No we don't need them.

src/application-base/ApplicationBase.jsx
mjhenkes: Yes

src/application-base/ApplicationBase.jsx
mjhenkes: Almost, In the next MVB for terra-application I think we'll remove `fitToParentIsDisabled ` So we may still be able to kill this div.

ryanthemanuel (approved):

dkasper-was-taken (approved):

src/terra-dev-site/doc/application/reference.1/contexts/ThemeContext.doc.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+<ThemedComponent description="This example pulls the current theme from the theme context and adds the theme class name to apply theme variables." isExpanded />
+
+## Context Value
+
+The ThemeContext provides an object with the following values:
+
+|Key Name|Type|Is Required|DefaultValue|Description|
+|---|---|---|---|---|
+|`className`|String|optional|undefined|The current application theme className. The default theme is indicated as undefined or empty string.|
```
tbiethman, line 25: > The default theme is indicated as undefined or empty string.

Can it really be either?

mjhenkes: by default it would be undefined if you passed nothing, but i suspect it's set to empty string in terra-dev-site

mjhenkes: So I guess brass tacks are that we don't control what the default theme is indicated as, but generally we expect it would be either undefined or empty-string. To have more assurance here we'd have to dictate how themes are included at the terra application layer.

tbiethman (approved):

PR #24, opened 2020-04-13
Add event listener to ApplicationBase for full stack testing

yuderekyu:
### Summary

Add hook `useFullStackTestOverrides` to ApplicationBase - this allow full stack testing to override locale for testing via dispatched event

### Deployment Link

https://terra-application-deployed-pr-24.herokuapp.com/

### Testing

Verified via wdio test that overrides locale to `pt`

### Additional Details

@cerner/terra


src/application-base/private/useFullStackTestOverrides.js
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from 'react';
+
+const FULL_STACK_OVERRIDE_EVENT = '[full-stack]override';
```
emilyrohrbough: This goes back to my question yesterday, would there be a use case we may want to use this test overrides event outside of full stack testing? Do we want this name to be specifically to full-stack? If so I think these changes look good.

yuderekyu: ah that's right. you thought of timezone as a potential use case at the component level, right? if we go generic, any ideas on the name? maybe `[test]override`?

yuderekyu: refactored to generic test override. namespaced the event as `applicationBase.testOverride` https://github.com/cerner/terra-application/pull/24/commits/ad362ccd4b397920fcfb4ffacd1c91c2d3c5b9fb

emilyrohrbough (approved):

ryanthemanuel (approved):

src/application-base/private/useTestOverrides.js
```diff
@@ -0,0 +1,33 @@
...
+const APPLICATION_BASE_OVERRIDE_EVENT = 'applicationBase.testOverride';
+
+/**
+ * Custom React hook that overrides locale via events. Leveraged by full stack testing to control deployed applications.
+ * @returns {Object} Contains the locale to override with.
+ */
+const useTestOverrides = () => {
+  const [localeOverride, setLocaleOverride] = useState();
+
+  const handleTestOverrides = (event) => {
```
tbiethman: I'd move this function definition into the useEffect block. There's no need to recreate with every update; after the first render, you're not going to do anything with it again.

src/application-base/private/useTestOverrides.js
```diff
@@ -0,0 +1,33 @@
...
+      setLocaleOverride(metaData.locale);
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener(APPLICATION_BASE_OVERRIDE_EVENT, handleTestOverrides);
+    return () => {
+      window.removeEventListener(APPLICATION_BASE_OVERRIDE_EVENT, handleTestOverrides);
+    };
+  }, []); // Empty array ensures we setup and teardown event listener once.
```
StephenEsser, line 28: We might need our resident hooks expert to double check the function references won't be an issue here. The `handleTestOverrides` will be redefined on each render, but the useEffect listener is only initialized once. 

cc: @tbiethman

tbiethman (approved):

src/application-base/private/useTestOverrides.js
mjhenkes: Could you also define it outside the useTestOverrides to not recreate it each update? If you passed the hooks to it...

src/application-base/private/useTestOverrides.js
tbiethman: It'd work fine, as the correct function instance would be captured in the arrow function at the time it is created. But as it is written, we WOULD create a new function every time the hook is executed and just do nothing with it. Not going to be a performance bottleneck or anything, but it's kind of confusing to look at.

src/application-base/private/useTestOverrides.js
yuderekyu: oops good catch, moved into useEffect https://github.com/cerner/terra-application/pull/24/commits/1be4958a1fca0926deb7379a4ea5aab5848bf38f

mjhenkes (approved):

tbiethman (approved):

PR #29, opened 2020-04-17
Disclosure stack

### LINKED ISSUE #16: Allow support Android back button navigation
# Feature Request
Add support for Android back button navigation. 

## Description
In our Android shell apps, pressing the back button will close the application when there is not a back action to take on the screen.  This will not be what users expect.  Create an API that can get triggered from the Android back button event in shell apps - either automatically or something the shell apps can call into.

If no action was taken, the native app will need to know so it can close the app.

This behavior was supported in legacy terra and will look like a defect if it stops working.

## @ Mentions
@mjhenkes, @tbiethman

dkschoonover:
### Summary
Track callback functions to close currently disclosed content.  Each disclosure will be registered to the same stack.  This will allow disclosures to be be closed by external events like the Android back button.

Closes #16 

### Deployment Link

https://terra-application-deployed-pr-29.herokuapp.com/

### Testing
Changes have been unit tested and functionally tested with WDIO.  I have plugged these changes into orion-application and tested an Android shell app.

### Additional Details

@cerner/terra


src/disclosure-manager/_DisclosureContainer.jsx
```diff
@@ -34,6 +35,14 @@ const DisclosureContainer = injectIntl(({ intl, children, navigationPromptResolu
 
   const defaultPromptOptions = useMemo(() => getUnsavedChangesPromptOptions(intl), [intl]);
 
+  useEffect(() => {
+    pushCallback(disclosureManager.goBack || disclosureManager.closeDisclosure);
+
+    return () => {
+      popCallback();
```
tbiethman: This implementation is making the assumption that the dismissal order will mirror the disclosure order. While this is true for any given instance of a DisclosureManager, this is not guaranteed to be true across multiple instances of DisclosureManager. With multiple DisclosureManagers all registering to the singleton _disclosureCallbackStack, there exists the possibility that something deep within the stack could be dismissed (think a slide panel being dismissed due to some async action while behind a presented modal). This would result in a dismissal stack that is inaccurate for the life of the remaining disclosures.

You might be able to work around this by beefing up the popCallback here such that it searches for and removes the instance of the function from deep within the "stack" rather than popping whatever is currently on top.

tbiethman: Similarly, if a SlidePanelManager were to disclose something (due to an async action) while presented behind a modal, the back button would dismiss the panel first before dismissing the modal. I don't see a way around this given the current architecture of things. 

To solve this problem, we would need to have a way at the framework level to detect different active "layers" of the framework and limit these types of actions (back button/escape key presses) to only the active layer. These are changes we need for other reasons as well, so we'll have to take this back button scenario into account when this work is done.

tbiethman: These scenarios would _also_ be a problem if multiple ModalManagers were presenting modals at the same time; however, we all know that there are more immediate problems with those scenarios, so the SlidePanelManager is a more realistic concern.

dkschoonover: That's a good point, this was assuming one instance of DisclosureManager.  I will rewrite with that taken into account.

dkschoonover: Updated the PR to remove the callback associated with the un-mounted container instead of the most recent disclosure.

src/disclosure-manager/_disclosureCallbacks.js
```diff
@@ -0,0 +1,24 @@
...
+const removeCallback = (callback) => {
+  disclosureCallbacks = disclosureCallbacks.filter(disclosureCallback => disclosureCallback !== callback);
+};
+
+const closeMostRecentDisclosure = () => {
+  if (!disclosureCallbacks.length) {
+    return false;
+  }
+
+  disclosureCallbacks[disclosureCallbacks.length - 1]();
```
tbiethman: Both goBack/closeDisclosure will return a Promise. The Promise will resolve if the dismissal happens, but it will be rejected if the user prevents the dismissal. I would return that promise here to give the caller the most flexibility in handling those different workflows.

 You could return a resolved promise directly if there's nothing in the stack. You could expose a `getStackSize` function or something if you think consumers of this will need to know that information.

dkschoonover: Updated to return a promise.  This will change a bit how the android integration will work.  Before I was planning on just blindly calling the dismiss function and notifying the Android shell if it returned false.

With these changes, "getActiveDisclosureCount" will be needed and will need to be checked before attempting to dismiss. If it returns 0 we can notify android instead of calling dismiss.

tbiethman:
@dkschoonover thank you for your work on this. I think this will work well for your short-term needs. Want to get some tests added so we can move forward with this?

dkschoonover:
Absolutely!  I will begin work on those now.

src/terra-dev-site/doc/application/reference.1/components/DisclosureManager.doc.mdx
```diff
@@ -0,0 +1,12 @@
+# DisclosureManager
+
+Manages application disclosures. See terra-disclosure-manager for general usage details.
+
+## Dismiss Callback Stack
+Content disclosed with terra-application's DisclosureManager can be dismissed from outside of a
```
dkschoonover: Let me know if there's a better place to put this information or if you want it documented at all.

tests/jest/disclosure-manager/disclosureCallbacks.test.js
```diff
@@ -0,0 +1,52 @@
...
+  removeCallback,
+  closeMostRecentDisclosure,
+  getActiveDisclosureCount,
+} from '../../../src/disclosure-manager/_disclosureCallbacks';
+
+const callback1 = () => 1;
+const callback2 = () => 2;
+const callback3 = () => 3;
+
+describe('disclosureCallbacks', () => {
```
dkschoonover, line 12: Not thrilled with these tests, but had to workaround not having the callback array directly exposed.

src/terra-dev-site/doc/application/reference.1/components/DisclosureManager.doc.mdx
tbiethman: I'd vote no docs at all. I don't want anyone consuming this outside of our internal usage.

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
```diff
@@ -81,4 +81,25 @@ Terra.describeViewports('SlidePanelManager', ['large'], () => {
...
+    browser.click('#root-component .disclose-large');
+    browser.waitForVisible('[class*="slide-group"] #DemoContainer-1 .global-close-disclosure');
+
+    browser.click('[class*="slide-group"] #DemoContainer-1 .disclose-tiny');
+    browser.waitForVisible('[class*="slide-group"] #DemoContainer-2 .global-close-disclosure');
+
+    Terra.validates.element('10. nested panel-reopened', { selector });
+  });
+
+  it('dismisses the nested modal using the global close', () => {
```
tbiethman:
```suggestion
  it('dismisses the nested panel using the global close', () => {
```

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
```diff
@@ -81,4 +81,25 @@ Terra.describeViewports('SlidePanelManager', ['large'], () => {
...
+    browser.click('[class*="slide-group"] #DemoContainer-2 .global-close-disclosure');
+    browser.waitForVisible('[class*="slide-group"] #DemoContainer-1 .global-close-disclosure');
+    Terra.validates.element('11. nested panel-global dismiss', { selector });
+  });
+
+  it('dismisses the nested modal using the global close', () => {
```
tbiethman:
```suggestion
  it('dismisses the nested panel using the global close', () => {
```

src/terra-dev-site/doc/application/reference.1/components/DisclosureManager.doc.mdx
dkschoonover: Removed https://github.com/cerner/terra-application/pull/29/commits/3071d812782e671283d07dabc9e7cbeefb5a4b6b

tests/wdio/slide-panel-manager/slide-panel-manager-spec.js
dkschoonover: Yikes, resolved here: https://github.com/cerner/terra-application/pull/29/commits/3071d812782e671283d07dabc9e7cbeefb5a4b6b

tbiethman (approved):

mjhenkes (approved):

src/disclosure-manager/_DisclosureContainer.jsx
```diff
@@ -34,6 +35,15 @@ const DisclosureContainer = injectIntl(({ intl, children, navigationPromptResolu
...
   const defaultPromptOptions = useMemo(() => getUnsavedChangesPromptOptions(intl), [intl]);
 
+  useEffect(() => {
+    const callback = disclosureManager.goBack || disclosureManager.closeDisclosure;
+    addCallback(callback);
+
+    return () => {
+      removeCallback(callback);
+    };
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
emilyrohrbough, line 45: Did we want to update this if disclosureManager changes?

src/disclosure-manager/_disclosureCallbacks.js
```diff
@@ -0,0 +1,26 @@
...
+
+const closeMostRecentDisclosure = () => {
+  if (!disclosureCallbacks.length) {
+    return Promise.resolve();
+  }
+
+  return disclosureCallbacks[disclosureCallbacks.length - 1]();
+};
+
+const getActiveDisclosureCount = () => (disclosureCallbacks.length);
```
emilyrohrbough, line 32: Out of curiosity, how or when do we expect this to be used?

src/terra-dev-site/test/modal-manager/DisclosureComponent.jsx
```diff
@@ -177,6 +182,7 @@ class DisclosureComponent extends React.Component {
         {disclosureManager && disclosureManager.goBack ? <button type="button" className="go-back" onClick={this.goBack}>Go Back</button> : null}
         {disclosureManager && disclosureManager.maximize ? <button type="button" className="maximize" onClick={this.maximize}>Maximize</button> : null}
         {disclosureManager && disclosureManager.minimize ? <button type="button" className="minimize" onClick={this.minimize}>Minimize</button> : null}
+        <button type="button" className="global-close-disclosure" onClick={closeMostRecentDisclosure}>Global Close</button>
```
emilyrohrbough: it might be nice to use `getActiveDisclosureCount` in this test to show the current number of disclosures that are open

src/disclosure-manager/_DisclosureContainer.jsx
dkschoonover: This is definitely worth discussing.  Re-running this if disclosure manager changes could introduce potential problems since we want to make sure that the array order is the disclosure order.  I was thinking that once disclosed these functions should never really change for the DisclosureContainer, but maybe that is a bad assumption.

One option that might make this more resilient would be to pass in a function here that references whatever disclosure manager is at the time of callback execution instead of passing in the disclosure manager functions directly.

Something like this:
```
const callback = () => (disclosureManager.goBack ? disclosureManager.goBack() : disclosureManager.closeDisclosure())
```

If we need to/want to support disclosureManager updates I think we would need some kind of "replaceCallback" function.  Any thoughts?

src/terra-dev-site/test/modal-manager/DisclosureComponent.jsx
dkschoonover: Agreed, will update.

src/disclosure-manager/_disclosureCallbacks.js
dkschoonover: When an event is triggered (for example, the android back button) the consumer can call this to help determine what action to take.  If there are disclosures, they would probably call closeMostRecentDisclosure.  If not, they would take some other action.

emilyrohrbough: can we add some document on the use case?

dkschoonover: Originally I had some documentation, but we decided to remove it (conversation is hidden above) since this isn't really intended to be used externally.

src/disclosure-manager/_DisclosureContainer.jsx
tbiethman: The `disclosureManager` values are generated once for each disclosed component and stored in state. The value should not be changing.

src/terra-dev-site/test/modal-manager/DisclosureComponent.jsx
dkschoonover: Update: a8d53bc2450685c620c74f2468cc89a0c6ea11df

This will reflect the count at the time of disclosure and does not update.

src/disclosure-manager/_disclosureCallbacks.js
```diff
@@ -0,0 +1,26 @@
+let disclosureCallbacks = [];
```
emilyrohrbough, line 1: @tbiethman should this get pushed under private API?

src/disclosure-manager/_disclosureCallbacks.js
emilyrohrbough: @tbiethman against in-line docs for developers?

emilyrohrbough (approved):

src/disclosure-manager/_disclosureCallbacks.js
tbiethman: I'm not opposed to inline docs. I just really don't want consumers getting ahold of this and using it in their workflows.

src/disclosure-manager/_disclosureCallbacks.js
tbiethman: I waffled on this. orion-application consumes it and imports it directly. I'm not sure if we want orion-application to be importing private APIs. By making it private, we'd be giving ourselves free-reign to make non-passive changes to it here and maybe breaking orion-application in the process.

I think it makes sense to keep it public (to keep us honest with passivity) but to hide it as much as humanly possible (i.e. no consumer facing docs in dev site).

src/disclosure-manager/_disclosureCallbacks.js
dkschoonover: Added some inline doc for the publicly exported functions: af5f40475a7c554a667353d1765d7b8c0956dba4

PR #32, opened 2020-04-22
Default error message

jmsv6d:
### Summary
Adding a default error message for the error boundary when there is none provided

### Deployment Link

https://terra-applic-default-er-qucogs.herokuapp.com/tests/terra-application/application-error-boundary/error-boundary-test

### Additional Details

@cerner/terra


ryanthemanuel (approved):

emilyrohrbough (approved):

benbcai (approved):

translations/en-AU.json
```diff
@@ -4,5 +4,6 @@
   "terraApplication.unsavedChangesPrompt.multiplePromptMessageIntro": "There are unsaved changes made to the following:",
   "terraApplication.unsavedChangesPrompt.multiplePromptMessageOutro": "Changes will be lost if you don't save them. How do you want to proceed?",
   "terraApplication.unsavedChangesPrompt.acceptButtonText": "Don't Save",
-  "terraApplication.unsavedChangesPrompt.rejectButtonText": "Continue Editing"
+  "terraApplication.unsavedChangesPrompt.rejectButtonText": "Continue Editing",
+  "terraApplication.errorBoundary.defaultErrorMessage": "We are currently unable to complete your request. Try again later, and contact your support team if this problem continues."
```
mjhenkes: I'm not sure these error boundary failures are the types of failures that will just work if you try again. Something massively went wrong in your application.

src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -76,14 +83,18 @@ class ApplicationErrorBoundary extends React.Component {
...
     if (activeError) {
+      let errorText = activeError.toString();
+      if (!activeError.message || activeError.message.length === 0) {
+        errorText = intl.formatMessage({ id: 'terraApplication.errorBoundary.defaultErrorMessage' });
+      }
       return (
         <StatusView
           variant="error"
-          message={activeError.toString()}
+          message={errorText}
```
mjhenkes, line 97: Could I get some more information about why we're changing the error boundary text? What cases are they being used for? I assumed they would be here in the case of major app failures. Are people seeing these a lot?

jmsv6d: If there's an error thrown without any error message, then the error view will display the error type without any indication to the user about what went wrong. This would provide some sort of information to the user in that case. This would likely only happen if a custom error was thrown  that didn't provide any message and was not caught anywhere.

mjhenkes: Have we seen instances where the thrown js exception doesn't have an error message? I don't mean to be difficult but I wondering what brought this change up?

ryanthemanuel: @mjhenkes, UX didn't like it.  Didn't think it was user friendly

StephenEsser (approved):

jenniferwurth:
@ryanthemanuel: Discussed with the UXFoundations group, and there are potentially two versions depending on the use case I'll detail out below.  Our traditional messaging when errors are present is to state what the issue is and how to resolve it. Typically and whenever possible, we would offer actions to the user to retry, log a ticket, or when no action is available to contact their system admin/IT support. Based on my conversations with Ryan this situation will only occur in non-prod environments where only an engineer or developer would encounter them, not a user (ie: clinician, revcycle staff, etc.) If this is the case where the user would know by using the error tracing what to correct. Our proposal is outlined below. please let me know if any assumptions are incorrect.

**Scenario 1: Error messaging is short & concise, less than 2 lines of text.**
- Status View Type: Error
- Icon: Error
- Title: Error (centered under error icon)
- Message: The system encountered an error: [insert error string here]. (left-aligned)

image



**Scenario 2: Error messaging is long & has lots of coding jargon.**
- Status View Type: Error
- Icon: Error
- Title: Error (centered under error icon)
- Message: The system encountered an error. (center-aligned)
- Display Show/Hide component centered below messaging. "Show [error log/string/tracking]" (or whatever appropriate description of what will be shown when selected. 
- Display full error string when the user selects "Show [error string/log/tracking]". Text should be left aligned. When error string is displayed, "Show [error log/string/tracking]" is replaced with "Hide"

image

cc: @neilpfeiffer @jeremyfuksa

ryanthemanuel:
@jmsv6d let's go with scenario 1 here for now.  I've logged https://github.com/cerner/terra-core/issues/3000 to let us explore doing something more along the lines of scenario 2 with the status view.

jeremyfuksa:
Since you added the new issue, can we remove the UX review label on this one?

mjhenkes (approved):

PR #35, opened 2020-04-30
Apply theme name to theme context

mjhenkes:
### Summary
This will use the theme config to apply the appropriate theme name to the theme context 

### Deployment Link

https://terra-applic-35.herokuapp.com/

### Testing
Jest

### Additional Details
I've also exported the theme provider here because I think it's required for testing theme changes in custom components.

Related PRs:
https://github.com/cerner/terra-toolkit-boneyard/pull/33
https://github.com/cerner/terra-framework/pull/1101

@cerner/terra

### DIFF
--- a/src/application-base/ApplicationBase.jsx
+++ b/src/application-base/ApplicationBase.jsx
@@ -17,9 +17,15 @@ import useTestOverrides from './private/useTestOverrides';
 
 import styles from './ApplicationBase.module.scss';
 
+/* global TERRA_THEME_CONFIG */
+
 const cx = classNames.bind(styles);
 
 const browserLocale = getBrowserLocale();
+
+// We only need to retrieve the root theme and root theme name once for the life of the application.
+const themeConfig = (typeof (TERRA_THEME_CONFIG) !== 'undefined') ? TERRA_THEME_CONFIG : undefined;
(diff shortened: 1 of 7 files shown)


emilyrohrbough (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -17,10 +17,16 @@ import useTestOverrides from './private/useTestOverrides';
...
 
+/* global TERRA_THEME_CONFIG */
+
 const cx = classNames.bind(styles);
 
 const browserLocale = getBrowserLocale();
 
+// We only need to retrieve the root theme and root theme name once for the life of the application.
+const themeConfig = (typeof (TERRA_THEME_CONFIG) !== 'undefined') ? TERRA_THEME_CONFIG : undefined;
+const rootThemeName = themeConfig?.theme ? themeConfig.theme : 'terra-default-theme';
```
mjhenkes, line 28: This logic, to read the config and apply the theme name when the theme class name is undefined, could be moved to terra-theme-context provider, but it felt weird to make these assumptions there. 

I tried to write a jest test for these, redefining the TERRA_THEME_CONFIG global, but i think this is untestable with these functions defined outside of the class. This is more efficient though because these methods are only run once.

tbiethman (approved):

StephenEsser (approved):

PR #36, opened 2020-05-06
onError callback added for errors caught in Error Boundary

### LINKED ISSUE #15: Provide a mechanism for consumers to provide a global error handler that gets executed on error boundaries
# Issue Description
Currently terra-application provides an error boundary, but when errors are caught they are suppressed from new relic.  New relic provides an API to log errors which could be used to forward errors if newrelic is present on the window.

https://docs.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/notice-error

lokesh-0813:
### Summary

With this change consumers can use the onError callback to log errors to a third party logger.

Closes #15 

### Deployment Link

### Testing

Added wdio's for onError callback in Application Base and Application Error Boundary Component.
### Additional Details

@cerner/terra


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 Changelog
 =========
+### Added
+* Updated error boundary to list errors in new relic.
```
saket2403: Move this to unreleased

lokesh-0813: Updated.

src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -72,6 +72,9 @@ class ApplicationErrorBoundary extends React.Component {
      * next update occurs.
      */
     this.errorRef.current = error;
+    if (error && window.newrelic) {
+      window.newrelic.noticeError(error);
```
emilyrohrbough: Is there any way to test this?

lokesh-0813: I was able to list the errors in new relic by running in production mode locally along with new-relic license key and terra-ui/terra-dev-site's application ID. The errors were registered for both the repos.

[Here](https://rpm.newrelic.com/accounts/1721539/browser/142381315/errors#?groupBy=errorMessage&tw_start=&tw_end=now&tw_offset=259200000) is a link to the registered JS errors in new-relic for terra-ui. If you are unable to see the errors change the time-picker to last 7 days. Errors that say "whoops" and "This is an error test"  was simulated locally in production mode.

ryanthemanuel:
I read @nickpith's post on the original issue and I like it.  I think to fix this issue, we should add a prop to the error boundary and application base called onError or something.  And then we execute that function instead of calling new relic explicitly.  That way we could change to something else in the future if needed.

We would then need to supply new relic as our method of logging errors in orion-application and terra-dev-site (https://github.com/cerner/terra-dev-site/blob/1ac49b74542ff1a6443f5de6601e965af5adced3/src/TerraDevSite.jsx#L20).

lokesh-0813:
> I read @nickpith's post on the original issue and I like it. I think to fix this issue, we should add a prop to the error boundary and application base called onError or something. And then we execute that function instead of calling new relic explicitly. That way we could change to something else in the future if needed.
> 
> We would then need to supply new relic as our method of logging errors in orion-application and terra-dev-site (https://github.com/cerner/terra-dev-site/blob/1ac49b74542ff1a6443f5de6601e965af5adced3/src/TerraDevSite.jsx#L20).

Yeah that makes sense. I have added the callbacks and removed newrelic api calls.
Updated here 1596939fddb165590419bc23e9bf9f3d293cb8a5

src/application-base/ApplicationBase.jsx
```diff
@@ -73,10 +73,14 @@ const propTypes = {
    * before continuing with the unload event, even if NavigationPrompts are present.
    */
   unloadPromptIsDisabled: PropTypes.bool,
+  /**
+   * A Callback for errrors caught by the Error Boundary.
+   */
+  onError: PropTypes.func,
```
mjhenkes, line 79: We need to be way more descriptive with this prop.

mjhenkes: Should this apply to all Application error boundaries when it's set here? Or should it be available in from terra-application to propagate to them?

ryanthemanuel:
Closing this for now since we're going with a different strategy

PR #47, opened 2020-06-17
Application Status Overlay

### LINKED ISSUE #30: Provide Application Page Status View API
# Feature Request

## Description
Similarly to how we provide the application overlay and overlay provider we should offer an application page status view and application page status view provider.

I imagine this status view should overlay content instead of be rendered amongst it. We can also take this opportunity to limit the api.
One status view per provider.

It could look something like this to consume.
```jsx
import React, { useState } from 'react';
import ApplicationPageStatus from 'terra-application/lib/application-page-status';
const ExampleComponent = () => {
  const [isError, setIsError] = useState(false);
  return (
    <div>
      <p>This component toggles an error state that lasts for 3 seconds.</p>
      <p>
        <button type="button" onClick={() => {
          setIsLoading(true);
          setTimeout(() => {
            setIsError(false);
          }, 3000);
        }}>
          Start Loading
        </button>
      </p>
      <ApplicationPageStatus type='Error' isOpen={isLoading} message='this page failed'/>
    </div>
  );
};
export default ExampleComponent;

```

avinashg1994:
### Summary

The following additions/changes have been made in this PR:

1. Added **ApplicationStatusOverlay** and **ApplicationStatusOverlayProvider** APIs.
2. Only one Status View can be rendered per ApplicationStatusOverlayProvider. If more than one ApplicationStatusOverlay is specified, then the last rendered gets the preference
3. Restricted the terra-status-view props to include only:
   - buttonAttrs
   - message
   - title
   - variant
4. Provided ApplicationStatusOverlayProvider through the following:
   - ApplicationBase
   - ApplicationNavigation
   - ModalManager
   - SlidePanelManager

Closes #30 

### Deployment Link

https://terra-applic-applicatio-8mn4zu.herokuapp.com/home

### Testing

Added and tested using Jest and WDIO tests


package.json
```diff
@@ -113,6 +113,7 @@
     "terra-status-view": "^4.10.0",
     "terra-theme-context": "^1.0.0",
     "terra-theme-provider": "^4.0.0",
+    "terra-button": "^3.39.0",
```
yuderekyu: should we be pulling this in? i thought terra-button was already a dependency of terra-status-view

tbiethman:
@avinashg1994 Thank you for putting this together. A couple things:

* I would recommend updating the base name for this component from ApplicationPageStatus to ApplicationStatusOverlay. I'm open to challenges here, but this _is_ an overlay. It's not transparent, but it overlays the other page content. And while the concept of a "Page" is something we plan on introducing in the future, I'm not sure I want to get that term out here now without the related components in place.

* Similarly, the status overlay should have similar accessibility considerations to the ApplicationLoadingOverlay. We need to inert the content behind the status view. Otherwise, users will be able to tab behind it and screen readers will access it. You can manipulate this in the demo if you're quick. We can't rely on consumers rendering nothing behind the overlay to account for this.

src/application-page-status/ApplicationPageStatusProvider.jsx
```diff
@@ -0,0 +1,87 @@
...
+      });
+    },
+  }), []);
+
+  const registeredPageStatusKeys = Object.keys(registeredPageStatus);
+  const registeredButtonAttrs = registeredPageStatusKeys.map(key => (registeredPageStatus[key] && registeredPageStatus[key].buttonAttrs));
+  const registeredMessage = registeredPageStatusKeys.map(key => (registeredPageStatus[key] && registeredPageStatus[key].message));
+  const registeredTitle = registeredPageStatusKeys.map(key => (registeredPageStatus[key] && registeredPageStatus[key].title));
+  const registeredVariant = registeredPageStatusKeys.map(key => (registeredPageStatus[key] && registeredPageStatus[key].variant));
+
```
tbiethman: These repeated maps seem inefficient. Would it make more sense to extract the data you'll use for presentation into a separate state key to limit the logic necessary to re-render?

src/application-page-status/ApplicationPageStatus.jsx
```diff
@@ -0,0 +1,71 @@
...
+      buttonAttrs,
+      message,
+      title,
+      variant,
+    });
+
+    return () => {
+      applicationPageStatus.hide(statusViewId);
+    };
+  }, [buttonAttrs, message, title, variant, id, applicationPageStatus]);
```
tbiethman: You may consider splitting the show/hide calls into separate effects. Really, the only time you want to hide the overlay is if it completely unmounts. You'd just need to make sure that your provider can appropriately handle multiple `show` calls for the same id without manipulating the registration order.

tbiethman:
```jsx
  useLayoutEffect(() => {
    applicationPageStatus.show(id.current, {
      buttonAttrs,
      message,
      title,
      variant,
    });
  }, [buttonAttrs, message, title, variant, id, applicationPageStatus]);

  useLayoutEffect(() => () => {
    applicationPageStatus.hide(id.current);
  }, []);
```

package.json
avinashg1994: Yes, terra-status-view does have terra-button as a dependency.
The only reason for adding terra-button as dependency here is due to the proptype definition [here](https://github.com/cerner/terra-application/pull/47/files#diff-cd1f5ff4f736d3f9cc1b643e78fe160eR14)

src/application-page-status/ApplicationPageStatus.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-application/pull/47/commits/108d8ce2f7edd49a0a770fdb02a18227e2239c77

src/application-page-status/ApplicationPageStatusProvider.jsx
avinashg1994: i agree, so i have made the necessary changes here : https://github.com/cerner/terra-application/pull/47/commits/108d8ce2f7edd49a0a770fdb02a18227e2239c77

avinashg1994:
> * I would recommend updating the base name for this component from ApplicationPageStatus to ApplicationStatusOverlay.

Renamed to ApplicationStatusOverlay and updated all references

> * Similarly, the status overlay should have similar accessibility considerations to the ApplicationLoadingOverlay. We need to inert the content behind the status view. Otherwise, users will be able to tab behind it and screen readers will access it.

Added the logic to make the background children elements inert when the status is overlaid

src/application-status-overlay/ApplicationStatusOverlay.jsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const ApplicationStatusOverlay = (props) => {
+  const {
+    buttonAttrs,
+    message,
+    title,
+    variant,
+  } = props;
+  const id = useRef(uuidv4());
```
tbiethman: Nitpicking here, but one convention I've used elsewhere is the addition of a 'Ref' suffix to variables that hold refs. I find it's a good reminder to access the `current` property on it. So this would ideally be `idRef`.

I'm realizing I _didn't_ do that for the ApplicationLoadingOverlay's id variable. We should update that one too :P

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
```diff
@@ -0,0 +1,119 @@
...
+        const newRegisteredStatusOverlay = { ...state };
+        delete newRegisteredStatusOverlay[key];
+
+        return newRegisteredStatusOverlay;
+      });
+    },
+  }), []);
+
+  const disableContainerChildrenFocus = () => {
+    if (refContainer.current && refContainer.current.querySelector('[data-status-overlay-container-content]')) {
```
tbiethman: You should put the ref on the container element [here](https://github.com/cerner/terra-application/pull/47/files#diff-ac0fedc9c031cc8f126738a12803ff63R106) and lose the queries.

src/application-status-overlay/ApplicationStatusOverlayProvider.module.scss
```diff
@@ -0,0 +1,20 @@
...
+    height: 100%;
+  }
+
+  .status-view {
+    background-color: #fff;
+    outline: none;
+    overflow: hidden;
+    position: absolute;
+    width: 100%;
+    z-index: 100;
```
tbiethman: This seems arbitrarily high. I think you could put an explicit z-index of 0 on `.container`, 1 on the `.status-view`, and 0 on the `.content-container` and guarantee the stacking order you're wanting.

tbiethman: FYI anyone who wants to learn about z-index magic: https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

Article is old as dirt but still accurate, as far as I can tell. Helped me visualize how some of this works.

src/application-status-overlay/ApplicationStatusOverlay.jsx
avinashg1994: Renamed ref variables for both ApplicationStatusOverlay and ApplicationLoadingOverlay here : https://github.com/cerner/terra-application/pull/47/commits/5e92dd7f93fe585a4883c6aa4734cb2859e0fa06

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
avinashg1994: Changed here : https://github.com/cerner/terra-application/pull/47/commits/5e92dd7f93fe585a4883c6aa4734cb2859e0fa06

src/application-status-overlay/ApplicationStatusOverlayProvider.module.scss
avinashg1994: The intention for the high z-index value was to account for user defined z-index values for the children elements.
But i agree that adding z-index of 0 to the `div` wrapping the children (container-content) does seem to address my concern much better.
Addressed here : https://github.com/cerner/terra-application/pull/47/commits/5e92dd7f93fe585a4883c6aa4734cb2859e0fa06

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
```diff
@@ -53,6 +68,23 @@ const ApplicationStatusOverlayProvider = ({ children, scrollRefCallback, ...cust
...
+    // eslint-disable-next-line no-prototype-builtins
+    if (!Element.prototype.hasOwnProperty('inert')) {
+      // IE10 throws an error if wicg-inert is imported too early, as wicg-inert tries to set an observer on document.body which may not exist on import
+      // eslint-disable-next-line global-require
+      require('wicg-inert/dist/inert');
+    }
+  });
+
+  useEffect(() => {
+    if (registeredStatusOverlayKeys.length > 0) {
```
yuderekyu: nit, we can simplify this to a length check
```suggestion
    if (registeredStatusOverlayKeys.length) {
```

avinashg1994: Changed here : https://github.com/cerner/terra-application/pull/47/commits/347793a8fbb9ff8781d9fce146c1f5a351097b4f

avinashg1994:
Added `background-color` for the Status Overlay and also made it themeable : https://github.com/cerner/terra-application/pull/47/commits/347793a8fbb9ff8781d9fce146c1f5a351097b4f

tbiethman (approved):

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
```diff
@@ -0,0 +1,123 @@
...
+  const enableContainerChildrenFocus = () => {
+    if (containerRef.current) {
+      containerRef.current.removeAttribute('inert');
+      containerRef.current.removeAttribute('aria-hidden');
+    }
+  };
+
+  const registeredStatusOverlayKeys = Object.keys(registeredStatusOverlay);
+  // In the event when multiple ApplicationStatusOverlay's are provided, the last rendered wins
+  const lastRegisteredStatusOverlayKey = (registeredStatusOverlayKeys.length > 0) && registeredStatusOverlayKeys[registeredStatusOverlayKeys.length - 1];
```
yuderekyu: same length comment here

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
```diff
@@ -0,0 +1,123 @@
...
+  }, [registeredStatusOverlayKeys]);
+
+  let className = cx('container');
+  if (customProps.className) {
+    className = [className, customProps.className].join(' ');
+  }
+
+  const theme = React.useContext(ThemeContext);
+
+  const statusView = (registeredStatusOverlayKeys.length > 0) && (
```
yuderekyu, line 95: same length comment here

yuderekyu (approved):

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-application/pull/47/commits/11d487988b685ac04b59cd6597291af14c490a2d

src/application-status-overlay/ApplicationStatusOverlayProvider.jsx
avinashg1994: The reason for not removing `> 0` check here is because without it when StatusOverlayApi is not provided it renders a `0` in the DOM.

mjhenkes (approved):

src/application-status-overlay/ApplicationStatusOverlay.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * The descriptive text, displayed under the title.
+   */
+  message: PropTypes.string,
+
+  /**
+   * The title displayed under the glyph. Variants contain default titles that can be overridden by this prop.
+   */
+  title: PropTypes.string,
```
neilpfeiffer: Pretty sure we also wanted to no longer expose the `title` in this new implementation based on UX higher specifications to meet consistency & design standards: 
https://github.com/cerner/terra-application/issues/30#issuecomment-637568575

cc: @jeremyfuksa

src/application-status-overlay/ApplicationStatusOverlay.jsx
```diff
@@ -0,0 +1,69 @@
...
+import PropTypes from 'prop-types';
+import Button from 'terra-button';
+import uuidv4 from 'uuid/v4';
+import ApplicationStatusOverlayContext from './ApplicationStatusOverlayContext';
+
+const propTypes = {
+  /**
+   * An array of objects containing terra-button properties. A key attribute is required for each object.
+   * This array is used to render buttons in the bottom section.
+   * Example:`[{ text: 'Button 1', key: 1, size: 'medium', variant: 'action', onClick: onClickFunction}]`
```
neilpfeiffer: Example needs to be updated - there is no longer `size` as a button prop and `action` button should not be recommended or suggested to be used here, only neutral or emphasis.
```suggestion
   * Example:`[{ text: 'Refresh Application', key: 1, variant: 'neutral', onClick: myRefreshFunction}]`
```

src/terra-dev-site/test/application-status-overlay/StatusOverlayAllProps.test.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import ApplicationStatusOverlay from '../../../../lib/application-status-overlay/ApplicationStatusOverlay';
+import ApplicationStatusOverlayProvider from '../../../../lib/application-status-overlay/ApplicationStatusOverlayProvider';
+
+const ApplicationStatusOverlayAllProps = () => {
+  const StatusViewButtons = [
+    {
+      text: 'Action 1',
+      key: 1,
+      size: 'medium',
```
neilpfeiffer: `size` is no longer a valid button prop, also remove on line 14

src/terra-dev-site/test/application-status-overlay/StatusOverlayAllProps.test.jsx
```diff
@@ -0,0 +1,25 @@
...
+    }, {
+      text: 'Action 2',
+      key: 2,
+      size: 'medium',
+    },
+  ];
+
+  return (
+    <ApplicationStatusOverlayProvider>
+      <ApplicationStatusOverlay buttonAttrs={StatusViewButtons} message="Status View with all props" title="All props test" variant="no-data" />
```
neilpfeiffer: with `title` prop removed, it will need to be removed here as well:
```suggestion
      <ApplicationStatusOverlay buttonAttrs={StatusViewButtons} message="Status View with all props" variant="no-data" />
```

PaulSolDev:
Functional review complete. No issues identified.

src/terra-dev-site/app/components.2/example/ApplicationStatusOverlayProps.jsx
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * The props table parser continues to have trouble detecting the propTypes of a component that
+ * returns `null`. The propType definition is duplicated here so that a table could be parsed for it.
+ */
+
+const propTypes = {
+  /**
+   * An array of objects containing terra-button properties. A key attribute is required for each object.
+   * This array is used to render buttons in the bottom section.
+   * Example:`[{ text: 'Button 1', key: 1, size: 'medium', variant: 'action', onClick: onClickFunction}]`
```
neilpfeiffer: Example needs to be updated - there is no longer `size` as a button prop and `action` button should not be recommended or suggested to be used here, only neutral or emphasis.
```suggestion
   * Example:`[{ text: 'Refresh Application', key: 1, variant: 'neutral', onClick: myRefreshFunction}]`
```

tests/jest/application-status-overlay/ApplicationStatusOverlay.test.jsx
```diff
@@ -0,0 +1,182 @@
...
+    expect(statusOverlayContextValue.hide.mock.calls.length).toBe(1);
+    expect(statusOverlayContextValue.hide.mock.calls[0][0]).toBe('test-id');
+  });
+
+  it('should render status view with the specified data', () => {
+    const StatusViewButtons = [
+      {
+        text: 'Action 1',
+        key: 1,
+        size: 'medium',
```
neilpfeiffer: `size` is no longer a valid button prop, also remove on line 67

src/application-status-overlay/ApplicationStatusOverlay.jsx
avinashg1994: Removed the `title` prop : https://github.com/cerner/terra-application/pull/47/commits/6fca8b18fef599c004d9efd588517068495dd45f

src/terra-dev-site/test/application-status-overlay/StatusOverlayAllProps.test.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-application/pull/47/commits/6fca8b18fef599c004d9efd588517068495dd45f

src/terra-dev-site/test/application-status-overlay/StatusOverlayAllProps.test.jsx
avinashg1994: Removed here : https://github.com/cerner/terra-application/pull/47/commits/62b43a41763ab55e834f07ce82e374ac33e0b409

src/terra-dev-site/app/components.2/example/ApplicationStatusOverlayProps.jsx
avinashg1994: Changed here : https://github.com/cerner/terra-application/pull/47/commits/62b43a41763ab55e834f07ce82e374ac33e0b409

tests/jest/application-status-overlay/ApplicationStatusOverlay.test.jsx
avinashg1994: Removed here : https://github.com/cerner/terra-application/pull/47/commits/62b43a41763ab55e834f07ce82e374ac33e0b409

src/application-status-overlay/ApplicationStatusOverlay.jsx
avinashg1994: Changed here : https://github.com/cerner/terra-application/pull/47/commits/62b43a41763ab55e834f07ce82e374ac33e0b409
Also, updated the **terra-status-view** documentation and references as well : https://github.com/cerner/terra-core/pull/3099

jeremyfuksa (approved):

PR #57, opened 2020-07-13
Update DisclosureContainer to allow custom registerDismissCheck implementations

tbiethman:
### Summary

The DisclosureContainer is wrapped around all components rendered by the DisclosureManager. It implements the DisclosureManager's registerDismissCheck API in order to provide NavigationPrompt integration. In doing so, the DisclosureContainer makes the registerDismissCheck API unusable by its children, even if the children desire a custom/different implementation.

I updated the DisclosureContainer to allow the `registerDismissCheck` function to be used by children. It turns out this was already documented as the shipped behavior, but the component mounting lifecycle resulted in this not working appropriately.

https://engineering.cerner.com/terra-ui/application/terra-application/contexts/disclosure-manager-context#a-note-on-registerdismisscheck

Note that providing a custom implementation will disable the default NavigationPrompt integration. This allows for maximum flexibility and aligns with the existing docs.

### Deployment Link

https://terra-applic-register-d-qyfgkx.herokuapp.com/

### Testing
I wrote new wdio tests that disable dismissal outright and validate the logic.

### Additional Details

@cerner/terra


src/disclosure-manager/_DisclosureContainer.jsx
```diff
@@ -45,26 +51,34 @@ const DisclosureContainer = injectIntl(({ intl, children, navigationPromptResolu
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
-    disclosureManager.registerDismissCheck(() => new Promise((resolve, reject) => {
-      if (!promptCheckpointRef.current) {
-        resolve();
-        return;
+    disclosureManager.registerDismissCheck(() => {
+      if (customRegisterDismissCheckRef.current) {
+        return customRegisterDismissCheckRef.current();
```
emilyrohrbough, line 56: Should this be 
```suggestion
        return customRegisterDismissCheckRef.current;
```
or should this execute and then return?

EDIT: wait this is a promise isn't it. NVM

src/terra-dev-site/test/modal-manager/CustomDismissCheck.test.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+import ModalManager from '../../../modal-manager';
+
+import DisclosureComponent from './DisclosureComponent';
+import styles from './ModalManagerCommon.test.module.scss';
+
+const cx = classNames.bind(styles);
+
+const CustomDismissCheck = () => (
+  <div role="main" className={cx('content-wrapper')}>
```
emilyrohrbough: Do we need the test wrapper?

tbiethman: As it turns out, I don't think so? Some prior tests had these, I removed them there and in these new tests.

src/disclosure-manager/_DisclosureContainer.jsx
tbiethman: Yep you got it, returning the Promise so that we can chain off it.

ryanthemanuel (approved):

emilyrohrbough (approved):

PR #58, opened 2020-07-14
Update travis-ci.yml

mjhenkes:
### Summary

This might not speed up performance, but lets align this with our other travis builds and add theme testing.

Closes #

### Deployment Link

### Additional Details

Basline:
Screen Shot 2020-07-15 at 9 46 48 AM

This PR:
Screen Shot 2020-07-15 at 9 55 21 AM

lol 2x more wdio tests in 1/2 the time!

@cerner/terra


tbiethman (approved):

tests/wdio/application-loading-overlay/application-loading-overlay-spec.js
```diff
@@ -2,6 +2,7 @@ Terra.describeViewports('ApplicationLoadingOverlay', ['large'], () => {
   describe('Loading overlay management', () => {
     before(() => {
       browser.url('/raw/tests/terra-application/application-loading-overlay/loading-overlay-test');
+      browser.moveToObject('#root', 0, 0);
```
mjhenkes, line 5: Funnily, when running against the static site the cursor is not set to root 0,0, and causes this test to fail, this doesn't happen when running against the serve static service.... i don't know why but this should fix it for both local and travis builds.

mjhenkes:
This PR is ready to go

.travis.yml
```diff
@@ -7,59 +7,69 @@ branches:
...
+      # Don't clean install
+      before_install: skip
+      # Rely on the cache for node_modules.. this won't work for any node modules in the packages dir but works for running wdio
+      install: skip
+      script:
+        # we start up our own standalone chrome instance because the Selenium Docker Service doesn't seem to work here.
+        - docker-compose up -d standalone-chrome-travis
+        # using the travis env section to set env variables seems to break cache
+        - SITE=$TRAVIS_BUILD_DIR/travis-build/default npm run wdio-default
+    - name: default-locale
```
emilyrohrbough, line 40: nit: default-theme-de-locale

docker-compose.yml
```diff
@@ -25,4 +25,14 @@ services:
...
     image: "${DOCKER_IMAGE}-test:${TAG:-latest}"
     depends_on:
-      - standalone-chrome
\ No newline at end of file
+      - standalone-chrome
+
+  standalone-chrome-travis:
+    image: selenium/standalone-chrome:3.14.0-helium
+    ports:
+      - 4444:4444
```
emilyrohrbough, line 27: Could we have updated the standalone-chrome service to use `4444:4444` to remove the need to add a new service?

docker-compose.yml
```diff
@@ -25,4 +25,14 @@ services:
     build: ./
     image: "${DOCKER_IMAGE}-test:${TAG:-latest}"
     depends_on:
-      - standalone-chrome
\ No newline at end of file
+      - standalone-chrome
```
emilyrohrbough: We can remove the test-ci service if I'm not mistaken

package.json
```diff
@@ -94,7 +96,12 @@
     "start-heroku": "tt-serve-static --port $PORT --site './build' ",
     "start-static": "npm run compile:prod && tt-serve-static --site ./build",
     "test": "npm run lint && npm run jest && npm run wdio",
-    "wdio": "tt-wdio --config node_modules/terra-toolkit/config/wdio/wdio.conf.js --locales=['en','de']"
+    "wdio-default": "wdio",
+    "wdio-locale": "LOCALE=de wdio",
+    "wdio-lowlight": "THEME=clinical-lowlight-theme wdio",
+    "wdio-fusion": "THEME=orion-fusion-theme wdio",
+    "wdio": "npm run wdio-default && npm run wdio-locale && npm run wdio-lowlight && npm run wdio-fusion",
+    "wdio-old": "tt-wdio --config node_modules/terra-toolkit/config/wdio/wdio.conf.js --locales=['en','de']"
```
emilyrohrbough: we could prob just delete this script

docker-compose.yml
mjhenkes: if we did it would mess with the local docker environment and bind standalone chrome to the local 4444 port... which would be annoying because those services aren't easily spun down when switching between dev envs

docker-compose.yml
mjhenkes: yeah, i can do that

package.json
mjhenkes: also true

.travis.yml
mjhenkes: we need to move this away from de which is why i named it default-locale... we're testing setting the default locale.... at least that was my thinking

mjhenkes:
for some reason my latest branch updates are showing up here?

wdio.conf.js
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const defaultWdioConfig = require('terra-toolkit/config/wdio/wdio.conf');
+
+const wdioConfig = defaultWdioConfig.config;
+
+const travis = process.env.TRAVIS;
+
+if (travis) {
+  wdioConfig.host = 'localhost';
```
emilyrohrbough, line 9: totally curios, if you set this value to be localhost but enable the selenium-docker-service, shouldn't this work correctly? I'm assuming you did try that..I'm just a little confused why this didn't work.

mjhenkes: that would be equivalent to disabling the CI env and running locally. It seemed like the selenium docker service had trouble starting up and would time out every time.

mjhenkes:
sometings messsed up with this PR, i'm going to close it and re-open it

emilyrohrbough (approved):

PR #61, opened 2020-07-16
Fix firefox tests due to out of range errors.

ryanthemanuel:
### Summary

Fix firefox tests due to out of range errors. Also prevent browser alerts on dirty data when we're not explicitly testing them in the navigation spec so that we don't cause timeout errors in selenium if there are problems

image


StephenEsser (approved):

benbcai (approved):

src/terra-dev-site/test/application-navigation/ApplicationNavigationDisabledPromptsTest.test.jsx
```diff
@@ -48,7 +50,7 @@ const ApplicationNavigationDisabledPromptsTest = () => {
   }];
 
   return (
-    <ApplicationBase locale="en">
+    <ApplicationBase locale={applicationIntl.locale}>
```
emilyrohrbough, line 53: I think this is only working out because we are rendered in terra-dev-site and the local is being provided by the root Application Base. That being said, I think this is fine since this is a test example.

emilyrohrbough (approved):

dkasper-was-taken (approved):

PR #72, opened 2020-08-19
ApplicationEventEmitter

### LINKED ISSUE #23: Application Event Emitter
# Feature Request

## Description
 Provide a user friendly event emitter to enable communication between otherwise unrelated components within an application. This should be a global api available to all components.

We could wrap dom events and clean up the API.
https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Overview_of_Events_and_Handlers

Or we could create a purely javascript pubsub.

There are open source options we should consider either providing directly or wrapping.
https://www.npmjs.com/package/events
https://www.npmjs.com/package/eventemitter3

benbcai:
### Summary
`ApplicationEventEmitter` can be used to create and emit custom events to facilitate communications between the emitting component (emitter) and the listening component (listener). The behavior of the event emitter is similar to the "pub/sub" model. When ApplicationEventEmitter emits an event, listeners that have registered to listen to this event can act upon receiving the event.

Added the lightweight [eventemitter3](https://github.com/primus/eventemitter3) dependency to leverage some performance gains.

Closes: #23 

### Deployment Link
https://terra-applic-event-emit-smhshu.herokuapp.com/application/terra-application/components/application-event-emitter

### Additional Details

@cerner/terra


src/terra-dev-site/app/components.2/ApplicationEventEmitter.app.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+```js
+import ApplicationEventEmitter from 'terra-application/lib/application-event-emitter';
+
+const eventEmitter = new ApplicationEventEmitter();
+let listenCount = 0;
+const listener = () => {listenCount += 1;}
+
+eventEmitter.on('event-name', listener);
+eventEmitter.on('event-name', listener);
```
mjhenkes: Should `eventEmitter.on('event-name', listener);` be listed twice in this example?

benbcai: I added twice here to illustrate that `off` removes all listeners registered for this event. I updated the `off` [API description and example](https://github.com/cerner/terra-application/pull/72/commits/845663e68f6bf6de6c0084464f88cbf26730bfce) to call out that it removes all listeners registered for this event. Do you think it's better to just list it once to minimize confusion?

package.json
```diff
@@ -104,6 +104,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "eventemitter3": "^4.0.4",
```
StephenEsser, line 107: Had to double check this wasn't a spoof package with a random number in there

src/terra-dev-site/app/components.2/ApplicationEventEmitter.app.mdx
mjhenkes: Yeah, i think it's a bit confusing, lets limit it to one.

mjhenkes:
@benbcai could we write up a guide walking through adding an event emitter to a react component and cleaning it up when the component goes away?

src/terra-dev-site/app/components.2/ApplicationEventEmitter.app.mdx
benbcai: Sounds good. https://github.com/cerner/terra-application/pull/72/commits/6b1d1ee20b96eebe53a4eb985b16b8fb7406f973

package.json
benbcai: Yeah, the package name should have been a bit more creative.

benbcai:
> @benbcai could we write up a guide walking through adding an event emitter to a react component and cleaning it up when the component goes away?

Yes, having one would be even more helpful for consumers.

src/terra-dev-site/app/components.2/ApplicationEventEmitter.app.mdx
```diff
@@ -0,0 +1,159 @@
+# Event Emitter
```
mjhenkes: I propose that event emitter should fall into the utilities section rather than components.

My reasoning is that components are react components and utilities are the extra stuff.

benbcai: That is a valid point. https://github.com/cerner/terra-application/pull/72/commits/30371e4bdb89d1e60da4cdf0f3e5dccbb9333544

Should we keep the name `ApplicationEventEmitter` or simply name it `EventEmitter` similar to us naming Logger instead of ApplicationLogger?

src/terra-dev-site/app/demo/EventEmitter.jsx
```diff
@@ -0,0 +1,50 @@
...
+class EventEmitterExample extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      counter: 0,
+    };
+
+    this.listener = this.listener.bind(this);
+    this.eventEmitter = new EventEmitter();
```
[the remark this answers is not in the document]
mjhenkes: I'm on the fence. If it was me developing an app, I'd do my best to create a functional component, but the example does seem a bit less clear.... I think i lean towards implementing this as a functional component and using hooks. Other thoughts?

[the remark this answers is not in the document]
benbcai: I thought about using a functional component with useEffect in this example similar to [this](https://reactjs.org/docs/hooks-effect.html#explanation-why-effects-run-on-each-update). I didn't go with that because this should only remove all listeners once when unmounting instead of after every render. There are certainly use cases where listeners may need to be added and removed as necessary after every render. This is a simple example illustrating the importance to remove all listeners when done.

mjhenkes (approved):
Looks good to me. I don't have strong feelings about changing the example to a functional component.

src/terra-dev-site/app/demo/EventEmitter.jsx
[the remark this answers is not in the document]
yuderekyu: the `useEffect` above will only re-trigger if the `eventEmitter` changes, signified by the array argument at the end 
```
}, [eventEmitter]);
```

not a big deal either way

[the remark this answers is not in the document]
benbcai: @yuderekyu  Ahh! I didn't know about the optimization option with the additional argument. I switched it and it works great. I can certainly change it to functional if others don't mind about it being less clear.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * Added
   * Added ApplicationStatusOverlay
   * Added ApplicationStatusOverlayProvider to ApplicationBase, ApplicationNavigation, ModalManager and SlidePanelManager
+  * Added EventEmitter
```
StephenEsser: This has snuck into a released section over time

src/terra-dev-site/app/demo/EventEmitter.jsx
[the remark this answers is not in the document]
StephenEsser: #teamHooks

src/terra-dev-site/app/demo/EventEmitter.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState, useEffect } from 'react';
+import EventEmitter from 'terra-application/lib/utils/event-emitter';
+
+const EventEmitterExample = () => {
+  const [counter, setCounter] = useState(0);
+  const [eventEmitter, setEventEmitter] = useState(() => new EventEmitter()); // eslint-disable-line no-unused-vars
```
yuderekyu: just curious, does react or eslint complain if we don't create the setEventEmitter variable?

src/terra-dev-site/app/demo/EventEmitter.jsx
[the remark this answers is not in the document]
benbcai: I've changed it to a function component using hooks.

CHANGELOG.md
benbcai: updated.

src/terra-dev-site/app/demo/EventEmitter.jsx
benbcai: I couldn't find any online reference or example for not having the setter. I tested it by removing it and it continues to work as expected. Neither react nor eslint complained. It appears that not including the setter just means that when useState returns the getter/setter pair, this would just ignore the setter that it returns.

yuderekyu: sweet. i can't think of any edge cases where we'd want to swap out the eventEmitter instance

yuderekyu: on second thought, b/c we remove the setter, we can also pass an empty array to useEffect to so we only the setup once
```
}, []);
```

yuderekyu (approved):

src/terra-dev-site/app/demo/EventEmitter.jsx
tbiethman: You could use `useRef` here instead. But thinking about that led me to wonder why we export a class at all. Can we export the singleton instance and call it good? Consumers wouldn't have to worry about this at all, just use the import directly. Or am I missing something?

```js
export default const EventEmitter = new EventEmitter3();
```

```jsx
import EventEmitter from 'terra-application/lib/util/EventEmitter';

const MyComponent = () => {
  React.useEffect(() => {
    const handler = () => {
      console.log('event happened');
    };

    EventEmitter.on('event', handler);

    return () => {
      EventEmitter.off('event', handler);
    };
  }, []);

  return <div />
};
```

We could also think about creating a custom hooks for some of the expected use cases. It would help to simply consumers' cleanup responsibilities, but the hooks wouldn't be super extensible or be able to handle dynamic use cases (start->stop->start listening).

```jsx
import EventEmitter from 'terra-application/lib/util/EventEmitter';

const useEventEmitter = (event, handler, deps) => {
  React.useEffect(() => {
    EventEmitter.on(event, handler);

    return () => {
      EventEmitter.off(event, handler);
    };
  }, deps);  
}

const MyComponent = () => {
  useEventEmitter('event', () => { console.log('event happened'); }, []);
  
  return <div />
};

const MyStateComponent = () => {
  const [state, setState] = React.useState('hi');

  useEventEmitter('event', () => { 
    console.log('event happened'); 
    console.log('current state: ' + state);
  }, [state]);
  
  return <div />
};

```

mjhenkes: @tbiethman's suggestion makes sense to me. (Now that i asked him about it 😄 )

mjhenkes: Though i think we should hold off on the custom hooks until we're sure about the pattern.

benbcai: Oh yes, just importing the singleton instance makes more sense. It is also easier to use for consumers. https://github.com/cerner/terra-application/pull/72/commits/4c35de80e0b282a1808de60894e5b33e086faccc

emilyrohrbough (approved):

src/terra-dev-site/app/demo/EventEmitter.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    const listener = () => {
+      setCounter(prevCounter => prevCounter + 1);
+    };
+
+    EventEmitter.on('increment-counter', listener);
+
+    return () => {
+      EventEmitter.removeAllListeners();
```
tbiethman: Can we change the example to turn off the specific event/handler? Anyone that copies this will remove all bound listeners (even the ones they didn't add) and be really confused.

```suggestion
      EventEmitter.off('increment-counter', listener);
```

benbcai: Fair enough: https://github.com/cerner/terra-application/pull/72/commits/7849a5af7ae295d8834691b474da133582ef304e

mjhenkes (approved):

tbiethman (approved):

PR #73, opened 2020-08-24
[terra-application] Reads out error boundary message

### LINKED ISSUE #56: Investigate our options for focus with terra-application-error-boundary
# Feature Request

Currently, it is indeterminate what happens to the active element when the error boundary shows (in chrome and firefox it goes to body, in ie it clears it).  We should figure out if there's something we can do better here.

lokesh-0813:
### Summary

Allows error boundary to read out error messages on screen readers.
~Focus is set on error boundary once rendered. ~

Closes #56 

### Deployment Link

### Testing

Tested in Chrome/Jaws and Safari/Voicover
### Additional Details

@cerner/terra


src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -92,10 +95,12 @@ class ApplicationErrorBoundary extends React.Component {
       const errorDetails = activeError.message.toString();
       const errorText = intl.formatMessage({ id: 'terraApplication.errorBoundary.defaultErrorMessage' }, { errorDetails });
       return (
-        <StatusView
-          variant="error"
-          message={errorText}
-        />
+        <div role="alert" id="applicationErrorBoundary" tabIndex="-1">
```
tbiethman: An `id` is probably safe, but using a `ref` would be more safe.

lokesh-0813: Switched to `ref`.

lokesh-0813:
With the current changes the browser adds a default focus-border around the application-error-boundary. Should we consider removing the border styles or wait till UX provides a value?

src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -92,10 +96,12 @@ class ApplicationErrorBoundary extends React.Component {
       const errorDetails = activeError.message.toString();
       const errorText = intl.formatMessage({ id: 'terraApplication.errorBoundary.defaultErrorMessage' }, { errorDetails });
       return (
-        <StatusView
-          variant="error"
-          message={errorText}
-        />
+        <div ref={this.errorBoundaryRef} role="alert" tabIndex="-1">
```
tbiethman: Is the "alert" role is appropriate here? It does ensure that the error text is "assertive" and that it's read when presented, but I'm not sure if there are other side effects.

src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -67,6 +67,10 @@ class ApplicationErrorBoundary extends React.Component {
 
   componentDidUpdate() {
     this.resetError();
+
+    if (this.errorBoundaryRef.current) {
+      this.errorBoundaryRef.current.focus();
```
tbiethman: Moving focus on every update is likely too often. The error boundary may be visible while the user interacts with other parts of the app (like the navigation header), and we don't want to move focus when those updates trickle down to the error boundary.

Can we limit this focus logic to just the initial presentation? I think _ideally_ we'd only move focus if the item that was previously focused is removed as part of the status view presentation, but I don't know if we can determine that reliably.

tbiethman: Thinking about this some more, I'm not sure if we need to be moving focus at all. Using the "alert" role might be enough to ensure the error is read and accessible. Setting focus to the error may make sense if the previously focused element was removed during the error presentation, but if not, the focus moving may be more obtrusive than we expect.

src/application-error-boundary/ApplicationErrorBoundary.jsx
ShettyAkarsh: Wrapping with div would add a default height. which would avoid error status to be displayed in the centre of the page when page level error boundary is implemented.  we might have to set the height of this div to 100% in that case.

src/application-error-boundary/ApplicationErrorBoundary.jsx
lokesh-0813: Removed focus in this commit 1b2a937569096398257177107586295f33a1a467

lokesh-0813: > Thinking about this some more, I'm not sure if we need to be moving focus at all. Using the "alert" role might be enough to ensure the error is read and accessible. Setting focus to the error may make sense if the previously focused element was removed during the error presentation, but if not, the focus moving may be more obtrusive than we expect.

That makes sense too.

src/application-error-boundary/ApplicationErrorBoundary.jsx
lokesh-0813: Removed the `DIV` and role is now a `customProp`

Changed here 1b2a937569096398257177107586295f33a1a467

ShettyAkarsh: I am not sure. but i had question like do we need to add this role attribute to status-view itself ? there might be this same accessibility issue in applications using status view. so which they have to fix it by passing the role attribute again similar to this. 

@tbiethman , @ryanthemanuel , @lokesh-0813  your thoughts on this  ?

tbiethman (approved):

mjhenkes (approved):

jeremyfuksa (approved):

PR #75, opened 2020-08-26
Lock down navigation-prompt-version

emilyrohrbough:
### Summary

Terra-application will need updated to handle the list formatting expected  with the latest designs of terra-notification-dialog/terra-navigation-prompt when there are multiple pending-actions in a notification dialog.

Currently there is caused a visual difference due to the native list styles.

See: https://github.com/cerner/terra-framework/pull/1105#issuecomment-676606484

### Deployment Link

https://terra-applic-lock-dialo-iurv0n.herokuapp.com/home

### Testing

Project tests.

### Additional Details

Updating changes to unlock this: https://github.com/cerner/terra-application/compare/unsaved-changes

@cerner/terra

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Changed
   * Updated `uuid` dependency to `v8.0.0`.
+  * Lock down terra-navigation-prompt to ensure MVB of terra-notification-dialog is consumed correctly
+
 * Fixed
   * Fix ApplicationStatusOverlayProvider contextType warning
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Lock down terra-navigation-prompt to ensure MVB of terra-notification-dialog is consumed correctly
+
 ## 1.31.1 - (Aug 19, 2020)
 
 * Changed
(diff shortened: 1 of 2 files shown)


tbiethman (approved):

package.json
```diff
@@ -113,7 +113,7 @@
     "terra-disclosure-manager": "^4.27.0",
     "terra-doc-template": "^2.2.0",
     "terra-modal-manager": "^6.20.0",
-    "terra-navigation-prompt": "^1.18.0",
+    "terra-navigation-prompt": "1.39.0",
```
StephenEsser, line 116: Does this need to be explicitly 1.39.0 or <= 1.39.0?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 
 * Changed
   * Updated `uuid` dependency to `v8.0.0`.
+  * Lock down terra-navigation-prompt to ensure MVB of terra-notification-dialog is consumed correctly
```
StephenEsser: This looks like it's in a released section

emilyrohrbough: My local version was out of date. It was correct initially 😓  I swear!

package.json
emilyrohrbough: It will alway pull the latests version (the current version is 1.39.0. ) with ^1.18.0 so IMO <=1.39.0 isn't needed since 1.39.0 is currently being used.

StephenEsser: > It will alway pull the latests version (the current version is 1.39.0. ) with ^1.18.0 

I don't think this is true if an application is specifying a specific version of terra-navigation-prompt. It might be worth a quick internal Github search for `terra-navigation-prompt` to double check if anyone is locking down versions. But I'd say minor version changes shouldn't be considered non-passive either way unless there's a specific scenario that causes breaking changes (React Context).

StephenEsser (approved):

PR #80, opened 2020-09-03
Setup publishing via travis-ci

mjhenkes:
### Summary

This PR sets up publishing via travis-ci similar to the changes that were made in the mono repos.

The scripts are derivations of the mono-repo versions, but not exactly the same. Ideally we'd be consolidating these scripts, but that's future Matt's problem.

Closes #

### Deployment Link

### Testing

This has been tested on this fork: https://github.com/mjhenkes/terra-application
https://travis-ci.com/github/mjhenkes/terra-application/builds/182614485

### Additional Details

@cerner/terra


package.json
```diff
@@ -100,7 +94,11 @@
     "wdio-locale": "LOCALE=de wdio",
     "wdio-lowlight": "THEME=clinical-lowlight-theme wdio",
     "wdio-fusion": "THEME=orion-fusion-theme wdio",
-    "wdio": "npm run wdio-default && npm run wdio-locale && npm run wdio-lowlight && npm run wdio-fusion"
+    "wdio": "npm run wdio-default && npm run wdio-locale && npm run wdio-lowlight && npm run wdio-fusion",
+    "version:major": "npm --no-git-tag-version version major",
+    "version:minor": "npm --no-git-tag-version version minor",
+    "version:patch": "npm --no-git-tag-version version patch",
+    "postversion": "node ./scripts/changelog-updater/index.js"
```
ryanthemanuel, line 102: Is this how npm works?  `postversion` does post of `version:<blah>`?

mjhenkes: postversion does a post of version, i added the major minor patch scripts to make it easier to not use git tagging

mjhenkes: Ideally i'd like to align this with the UI that lerna has to select a version, but this is the quick way to get our releases aligned.

mjhenkes: postversion is documented here https://docs.npmjs.com/cli/version

scripts/release/release.js
```diff
@@ -0,0 +1,53 @@
...
+
+const release = async () => {
+  if (await isPublished()) {
+    console.log('Nothing to publish');
+    process.exit();
+  }
+
+  setupGit();
+
+  const tag = `v${packageJson.version}`;
```
mjhenkes, line 42: tags are pushed following npm's tag format `v{version}`

tbiethman (approved):

scripts/release/release.js
```diff
@@ -0,0 +1,53 @@
...
+
+  try {
+    execSync('npm publish');
+  } catch (error) {
+    // console.log(error);
+    process.exit(1);
+  }
+  execSync(`git tag -a ${tag} -m "${tag}"`);
+  execSync('git push origin --tags');
```
emilyrohrbough, line 46: We should only tag & push this if the publish was successful. I know we have process.exit above, but why not add this in the try block as well?

package.json
emilyrohrbough: > Ideally i'd like to align this with the UI that lerna has to select a version, but this is the quick way to get our releases aligned.

Could you link the lerna UI? I'm curious about our own version scripts instead of using lerna's version command

scripts/release/release.js
mjhenkes: I'm not sure under what circumstances tagging fails and if it does i don't think theres harm in pushing no new tags. I figured publishing was the greater risk.

mjhenkes: I guess, i don't know if we need this try catch at all, i'll test

package.json
mjhenkes: this is the UI i'm referencing: 
Screen Shot 2020-09-03 at 12 16 56 PM

scripts/release/release.js
mjhenkes: I added better error handling

scripts/release/release.js
```diff
@@ -0,0 +1,52 @@
+#! /usr/bin/env node
+/* eslint-disable no-console */
+// eslint-disable-next-line import/no-extraneous-dependencies
+const pacote = require('pacote');
+const { execSync } = require('child_process');
+const { exit } = require('process');
+const packageJson = require('../../package.json');
+
+// If on travis setup git to be able to push the tags
+const setupGit = () => {
```
emilyrohrbough, line 10: Should this thrown an error if not in Travis ENV? or assuming this is supporting locale releases, should this run the prepublishOnly script, `npm whoami && check-installed-dependencies.

I'm wondering if this should be a promise to ensure git setup is correct / validate before running publish

mjhenkes: if the git setup fails at any point it will exit the script since it's wrapped in the async `release` function. I'm also fine with (people who have privs) still having the option to release locally if there is some type of emergency situation.

emilyrohrbough (approved):

PR #86, opened 2020-09-15
[v2] Initial updates to ApplicationBase

tbiethman:
### Summary

With terra-application@2.0, more structural components are being defined between ApplicationBase and the application's content. This initial PR removes features from ApplicationBase that are better suited for implementation within these new components. As a result, ApplicationBase becomes a much lighter and more reusable within our various embedded application architectures.

PRs for these new structural components are forthcoming. In the meantime, I've disabled the WDIO tests and left the dev-site documentation as-is until more of the pieces have landed. I'm trying to break these PRs up so that they are a bit more readable, but let me know if it's making things less clear.

### DIFF
--- a/src/application-base/ApplicationBase.jsx
+++ b/src/application-base/ApplicationBase.jsx
@@ -18,6 +18,8 @@ import { NavigationPromptCheckpoint } from '../navigation-prompt';
 import getBrowserLocale from './private/getBrowserLocale';
 import useTestOverrides from './private/useTestOverrides';
 
+import './private/initializeInert';
+
 import styles from './ApplicationBase.module.scss';
 
 const cx = classNames.bind(styles);
--- a/src/application-base/ApplicationBase.jsx
+++ b/src/application-base/ApplicationBase.jsx
@@ -1,27 +1,21 @@
 /* global TERRA_THEME_CONFIG */
 
 import React, {
-  useRef, useEffect, Suspense, useMemo,
+  useRef, useEffect, useMemo,
 } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames/bind';
 import Base from 'terra-base';
 import ThemeProvider from 'terra-theme-provider';
 import { ActiveBreakpointProvider } from 'terra-breakpoints';
 import ThemeContextProvider from 'terra-theme-context/lib/ThemeContextProvider';
 
-import ApplicationErrorBoundary from '../application-error-boundary';
 import { ApplicationIntlProvider } from '../application-intl';
-import ApplicationLoadingOverlay, { ApplicationLoadingOverlayProvider } from '../application-loading-overlay';
-import { ApplicationStatusOverlayProvider } from '../application-status-overlay';
 import { NavigationPromptCheckpoint } from '../navigation-prompt';
+
 import getBrowserLocale from './private/getBrowserLocale';
 import useTestOverrides from './private/useTestOverrides';
 
-import styles from './ApplicationBase.module.scss';
-
-const cx = classNames.bind(styles);
-
+// We determine the browser's default locale to be used when no locale is provided to ApplicationBase.
 const browserLocale = getBrowserLocale();
 
 // We only need to retrieve the root theme and root theme name once for the life of the application.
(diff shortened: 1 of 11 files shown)


src/application-base/ApplicationBase.jsx
```diff
@@ -105,50 +80,26 @@ const ApplicationBase = ({
...
-      <ThemeProvider
-        themeName={themeName}
-      >
-        <ThemeContextProvider theme={theme}>
-          <Base
-            customMessages={customTranslatedMessages}
-            translationsLoadingPlaceholder={translationsLoadingPlaceholder}
-            locale={localeOverride || locale || browserLocale}
-          >
-            <ApplicationErrorBoundary>
```
nickpith, line 131: Why remove the error boundary? That feels like a pretty core thing that even embedded apps may want or even other use cases like apps transitioning from legacy architectures that mix legacy UI frameworks with Terra components mounted on the page instead of a single React root.

tbiethman: That's a good point, those are definitely scenarios we want to support. But to be clear, the expectation is not that consumers of the framework will now need to provide their own error boundaries. Error boundaries will be provided by the framework and integrated within some of the new components being added to v2. I will call out where these are being added in future PRs to ensure we're on the same page.

nickpith: Ok I'll look for that. My main interest here is that apps transitioning from legacy architectures that are mounting React components onto DOM in existing pages can use `ApplicationBase` as a bootstrap for all the core things a single page app may have so that they can easily transition these separately mounted components into a single React root down the road without rework.

src/application-base/ApplicationBase.jsx
```diff
@@ -77,10 +44,18 @@ const propTypes = {
...
-  locale, customTranslatedMessages, translationsLoadingPlaceholder, themeName, fitToParentIsDisabled, children, unloadPromptIsDisabled,
+  locale, themeName, children, unloadPromptIsDisabled,
 }) => {
   const registeredPromptsRef = useRef();
 
+  const { localeOverride } = useTestOverrides(); // Allows us to test deployed applications in different locales.
+
+  const theme = useMemo(() => ({
+    // If the theme class name is undefined or an empty string, that indicates we have the root theme and should apply the root theme name.
+    name: themeName || rootThemeName,
```
emilyrohrbough, line 56: Should we expect the default theme name to come from the theme configuration? The fallback on this is a little weird since this could come from three places.

src/application-base/ApplicationBase.jsx
```diff
@@ -77,10 +44,18 @@ const propTypes = {
...
+    className: themeName,
```
emilyrohrbough, line 57: Should this fallback to `themeConfig?.theme`?

src/application-base/ApplicationBase.jsx
mjhenkes: `rootThemeName` is either the 'theme' field from theme config, or 'terra-default-theme' if it's not set. When we first setup the theme config, `theme` was and still is optional, so this gives a name to the default theme.

src/application-base/ApplicationBase.jsx
mjhenkes: it could, but themeName is either null, or undefined, or it should be filled out here. We could be more explict but this does work. This logic hasn't changed from terra-application v1, it's just moved.

src/application-base/ApplicationBase.jsx
```diff
@@ -38,36 +31,10 @@ const propTypes = {
...
-  /**
-   * The component to render while the translation files are being retrieved.
-   * NOTE: Absolutely no locale-dependent logic should be
-   * utilized in this placeholder.
-   */
-  translationsLoadingPlaceholder: PropTypes.node,
   /**
    * The name of the theme to apply to the application using terra-theme-provider.
    */
   themeName: PropTypes.string,
```
mjhenkes, line 38: I kind of wish this theme-name related to the name of the theme and not the css class to apply and application base would be smart enough to set the theme class to null if the lowlight theme is the default theme and you pass in the lowlight name. Right now you'd get the lowlight class name applied regardless. It still works fine, but kind of annoys me.

src/application-base/ApplicationBase.module.scss
```diff
@@ -1,9 +0,0 @@
-:local {
-  .application-base {
-    position: relative;
-  }
-
-  .application-base.fill {
-    height: 100%;
```
mjhenkes, line 7: 100% would get applied in the session provider now? or the application container?

tbiethman: The ApplicationContainer would be the first structural component to render to the DOM and define that height. I'd expect a SessionProvider to be either rendered within the ApplicationContainer, or to render and ApplicationContainer itself. Haven't decided yet.

src/application-base/ApplicationBase.jsx
```diff
@@ -40,36 +32,10 @@ const propTypes = {
...
```
ShettyAkarsh, line 38: should this theme name prop should accept only the list of theme class name which terra support as of now like do we need to do that validation here ?. i know that this is not the correct pr for doing that. but just a thought.

mjhenkes (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -107,50 +81,26 @@ const ApplicationBase = ({
...
-                </ActiveBreakpointProvider>
-              </ApplicationIntlProvider>
-            </ApplicationErrorBoundary>
-          </Base>
-        </ThemeContextProvider>
-      </ThemeProvider>
-    </div>
+    <ThemeProvider themeName={themeName}>
+      <ThemeContextProvider theme={theme}>
+        <Base locale={localeOverride || locale || browserLocale}>
```
emilyrohrbough, line 89: This will also need this change: https://github.com/cerner/terra-application/pull/95

emilyrohrbough (approved):

Issue #63, opened 2020-07-29
[spike] terra-application v3.0


tbiethman:
# Feature Request

## Description

The terra-application package will be adding features to further define the application architecture and simplify existing workflows and APIs. This issue details the new and changing functionality planned for the ~~v2.0~~v3.0 release. This is a living document and may change.

### New Features

New features that will be introduced include:

* `Page` - A component  defining consistent UX for application content and promoting application composition
  - Will have first-class APIs for page-level action controls, notifications, and progressive Page disclosure
  - Will integrate with existing ApplicationLoadingOverlay, ApplicationStatusOverlay, error boundaries, navigation prompts, etc.
  - Related components will provide optional side navigation and Workspace integration
* `Modals` - A simplified modal presentation component offered as a replacement for the ModalManager.
  - Will feature a declarative API to maintain Context provider expectations of the presenting component.
  - Will support additional Modals presented from a presented Modal (i.e. stacked modals).
* `ApplicationConcept` - A Context-based communication system allowing for the consistent presentation of Application-wide concepts (e.g. demographics information).
* `BlockingOverlay` - A new overlay component used to block user input for the entire application.
  - Will be useful for handling long write workflows where user navigation is not allowed.
  - Will feature UX patterns to minimize flashing
* `Secondary Navigation` - A component providing secondary navigation within a sidebar.
  - Will be useful for partitioning content further beyond the primary navigation provided by ApplicationNavigation.
* `Workspace` see https://github.com/cerner/terra-application/issues/85

### Functional Differences

Existing features are planned to change in the following ways:

* `ApplicationNavigation`
  - This component will move to a more opinionated API for its children to allow for the persisted mounting of each navigable child
  - Children are expected to implement Page (or related) components.
  - Component may be renamed to better align with new architecture concepts.
* `ModalManager`
  - We will attempt to uplift this component with the new Modal component provided by ~~v2~~v3. 
  - Regardless, the ModalManager will be **deprecated** in favor of direct Modal usage.

### New Dependencies

The following dependencies will be updated:

* `react-intl` - From v2.9 to current
  - Will allow for usage of new react-intl APIs and Context consumption throughout the framework

### Testing

Additional mock components will be created to aid developers' test creation.

* `MockApplicationBase` - A generic container for Jest-level test mounts that satisfies Context expectations.

### Existing Issues

Existing issues that will be corrected in this major release include:

- #62 
- #68

PR #87, opened 2020-09-16
[v2] Migrating terra-application-navigation from terra-framework

tbiethman:
### Summary

The ApplicationNavigation component is being migrated into terra-application to prepare for its alignment with v2's new architecture. This PR is a port of the previous component package's implementation and jest tests. The only changes include updates to the theme variable names to make them differ from those in the existing package, and the addition of dependencies for the migrated code. Wdio tests were not ported, as they will likely be rewritten as the component changes.


src/application-navigation/ApplicationNavigation.jsx
```diff
@@ -2,17 +2,18 @@ import React, {
...
 import ApplicationErrorBoundary from '../application-error-boundary';
 import ApplicationLoadingOverlay, { ApplicationLoadingOverlayProvider } from '../application-loading-overlay';
 import { ApplicationStatusOverlayProvider } from '../application-status-overlay';
 import { NavigationPromptCheckpoint, navigationPromptResolutionOptionsShape, getUnsavedChangesPromptOptions } from '../navigation-prompt';
 import { ApplicationIntlContext } from '../application-intl';
 
+import TerraApplicationNavigation from './terra-application-navigation/ApplicationNavigation';
+import {
+  titleConfigPropType, navigationItemsPropType, extensionItemsPropType, utilityItemsPropType, userConfigPropType,
+} from './terra-application-navigation/utils/propTypes';
```
emilyrohrbough, line 15: Should the terra namespace be dropped? Looking at the styles, they are weird seeing `terra-application-application`

I was also wondering if the structure should be flattened to move the terra-application-navigation contents up a dir level?
```
src/
    application-navigation/
        ApplicationNavigation.jsx
        index.js
        NavigationLayout (instead of terra-application-navigation/ApplicationNavigation)
```

src/application-navigation/terra-application-navigation/ApplicationNavigation.module.scss
```diff
@@ -0,0 +1,84 @@
+@import '../../breakpoints/media-queries';
+
```
emilyrohrbough, line 2: Does the terra-application-navigation have themes imports for all of the scss files?

src/application-navigation/terra-application-navigation/common/PopupCount.module.scss
```diff
@@ -0,0 +1,23 @@
+:local {
```
emilyrohrbough, line 1: Does this have themes?

src/application-navigation/terra-application-navigation/common/_PopupMenu.jsx
```diff
@@ -0,0 +1,217 @@
...
+};
+
+const defaultProps = {
+  menuItems: [],
+  isHeightBounded: false,
+  showSelections: false,
+  role: 'listbox',
+};
+
+const PopupMenu = ({
```
emilyrohrbough, line 104: I know this is a migration from terra-framework, but I'm wondering if we should add a high level explanation on more complicated packages that have multiple sub-components in it's implementation

src/application-navigation/terra-application-navigation/tabs/_TabCount.jsx
```diff
@@ -0,0 +1,54 @@
...
+const TabCount = ({ value, isRollup, intl }) => {
+  const countRef = useRef();
+
+  useAnimatedCount(countRef, value);
+
+  let validatedValue = value;
+  if (isRollup) {
+    validatedValue = intl.formatMessage({ id: 'Terra.applicationNavigation.notifications.new' });
+  } else if (value >= 999) {
+    validatedValue = '99+';
```
emilyrohrbough, line 36: Should this be 999+ like https://github.com/cerner/terra-application/pull/87/files#diff-31daef00e0f7bfe3c5f4c624e2a8575fR24?

themes/clinical-lowlight-theme/clinical-lowlight-theme.scss
```diff
@@ -0,0 +1,156 @@
+:global {
```
emilyrohrbough, line 1: Should this be converted or will there be a follow up pr to convert these to be theme imports?

tbiethman: If there's any modernizing to do for terra-application-navigation, I'd prefer to come back around with a follow-up PR.

src/application-navigation/terra-application-navigation/tabs/_TabCount.jsx
tbiethman: Hmm this definitely feels off, but I'm not sure whether we should update the check or update the string. @dkasper-was-taken do you remember?

src/application-navigation/ApplicationNavigation.jsx
tbiethman: I will move things around quite a bit with the introduction of the ApplicationContainer in my next PR. I wanted to get this code landed here first because there's so much of it. A bit of a half-step I know, but I think it'll make more sense to reorganize this in a followup.

I believe we need the `--terra-application` prefix to align with the naming rules, but I agree that seeing application twice in there is awkward. I can update these again when I come up with a better name for this thing.

src/application-navigation/terra-application-navigation/tabs/_TabCount.jsx
dkasper-was-taken: I think the if check portion is likely what is off. A navigation tab has a width restriction for fitting 99+. I feel like we fixed this elsewhere, but may have missed this one.

dkasper-was-taken (approved):

mjhenkes (approved):

src/application-navigation/terra-application-navigation/common/_PopupMenu.jsx
tbiethman: I do believe this would be helpful. I can take a stab at this when I make my changes to the Navigation component to support our new architecture.

emilyrohrbough (approved):
Can this be linked to https://github.com/cerner/terra-application/issues/63?

src/application-navigation/terra-application-navigation/ApplicationNavigation.module.scss
```diff
@@ -0,0 +1,84 @@
...
+    position: relative;
+    width: 100%;
+  }
+
+  .drawer-menu-container {
+    background-color: var(--terra-application-application-navigation-drawer-background-color, #f2f3f4);
+    height: 100%;
+    position: relative;
+    top: 0;
+    width: var(--terra-application-application-navigation-drawer-default-width, calc(100% - 3.571rem));
```
ShettyAkarsh, line 32: i believe this theme variables name be updated later.

src/application-navigation/terra-application-navigation/tabs/_TabCount.jsx
tbiethman: I updated this, and fixing in terra-framework with https://github.com/cerner/terra-framework/pull/1259

PR #95, opened 2020-09-28
Opt into throwing error if intl data fails to load

emilyrohrbough:
### Summary

Consume changes made in https://github.com/cerner/terra-core/pull/3173 to thrown an error if intl data fails to load.

### Deployment Link

### Testing

- unit & manual testing in https://github.com/cerner/terra-core/pull/3173

### DIFF
--- a/src/application-base/ApplicationBase.jsx
+++ b/src/application-base/ApplicationBase.jsx
@@ -18,6 +18,8 @@ import { NavigationPromptCheckpoint } from '../navigation-prompt';
 import getBrowserLocale from './private/getBrowserLocale';
 import useTestOverrides from './private/useTestOverrides';
 
+import './private/initializeInert';
+
 import styles from './ApplicationBase.module.scss';
 
 const cx = classNames.bind(styles);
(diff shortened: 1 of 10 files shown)


mjhenkes (approved):

tbiethman (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -119,12 +119,11 @@ const ApplicationBase = ({
...
   return (
     <div data-terra-application-base className={cx('application-base', { fill: !fitToParentIsDisabled })}>
-      <ThemeProvider
-        themeName={themeName}
-      >
+      <ThemeProvider themeName={themeName}>
         <ThemeContextProvider theme={theme}>
           <Base
             customMessages={customTranslatedMessages}
+            throwOnI18nLoadError
```
ShettyAkarsh, line 126: should this be caught by application error boundary ? i know currently base is not wrapped by error boundary. but with introduction of this do we need to wrap base also ?

emilyrohrbough: I originally had the same thought. The application error boundary relies on translations so we would have to introduce another error boundary for this level. I talked offline with @tbiethman and @mjhenkes  about this and we felt if a failure occurs at this level, there isn't much we can do to save the application since locales/translations data is needed for the app to render and function as expected.

PR #124, opened 2021-01-12
[V2] react-intl v5 upgrade

### LINKED ISSUE #111: Upgrade to react-intl v5
# Feature Request
As part of [terra-application v2](https://github.com/cerner/terra-application/issues/63), react-intl needs to be upgraded to v5. As part of this upgrade terra-base and terra-i18n will be deprecated with their functionality moving to terra-application-base.

## Description

The work for this has been started here: https://github.com/cerner/terra-application/tree/react-intl-take-2
There has also been work on updating the loaders here: https://github.com/cerner/terra-aggregate-translations 
We might want to discontinue generating the loader file and dynamically import locale data in terra-application-base as well. This would significantly simplify the logic around loading the locales.

jmsv6d:
### Summary
Moved i18n logic from terra-i18n and terra-base to be encapsulated in terra-application-base. Locale data loaders that were being generated by terra-aggregate-translations has been hard coded now and logic simplified to utilize a promise api. The biggest change required by `react-intl` v5 is an update to our polyfills, `react-intl` also no longer exports locale data. It requires the `Intl.PluralRules` and `Intl.RelativeTimeFormat` apis in addition to the one previously required. The polyfill `intl` that we have been utilizing supports `Intl.PluralRules` but not `Intl.RelativeTimeFormat`. `formatjs` provides polyfills for each of the `Intl` apis that it requires, but does not supply the locale data for all of the locales that we support. So, I continued to utilize the `intl` polyfill for the api's that it supplies and the `formatjs` polyfill for `Intl.RelativeTimeFormat`. 

Closes #111

### Deployment Link

### Additional Details

@cerner/terra


package.json
```diff
@@ -167,13 +170,12 @@
...
     "react-dom": "^16.13.1",
-    "react-intl": "^2.9.0",
+    "react-intl": "^5.8.2",
     "regenerator-runtime": "^0.13.5",
     "stylelint": "^10.0.1",
     "stylelint-config-terra": "^3.3.0",
     "terra-collapsible-menu-view": "^6.29.0",
-    "terra-dev-site": "^6.15.0",
-    "terra-enzyme-intl": "^3.3.0",
+    "terra-dev-site": "git+https://github.com/cerner/terra-dev-site.git#react-intl-upgrade",
```
mjhenkes: You're probably better off removing this dependency for now. When/if we add back terra-dev-site it'll have to be v7 and even then i'll be a weird circular dependency.

jmsv6d: Removed. I'd mostly had it in there to be able to test out my changes.

src/application-intl/ApplicationIntlContext.jsx
```diff
@@ -1,9 +0,0 @@
-import { createContext } from 'react';
-import { intlShape } from 'react-intl';
-
-const ApplicationIntlContext = createContext();
```
tbiethman: Does react-intl export their actual Context value, or just the useIntl custom hook? If it's just the hook, we might want to keep this around for the straggler Class-based components that can't use the hook.

tbiethman (approved):

src/application-base/ApplicationBase.jsx
```diff
@@ -49,9 +50,22 @@ const ApplicationBase = ({
...
 
   const { localeOverride } = useTestOverrides(); // Allows us to test deployed applications in different locales.
 
+  const finalLocale = localeOverride || locale || browserLocale;
+
+  useEffect(() => {
+    if (finalLocale !== undefined) {
+      i18nLoader(finalLocale).then(translationsModule => setMessages(translationsModule.default)).catch((error) => {
+        // eslint-disable-next-line no-console
+        console.error(error);
```
mjhenkes, line 64: Use the terra-application logger here:
https://engineering.cerner.com/terra-ui/application/terra-application/utilities/logger

src/application-base/private/translationsLoaders.js
```diff
@@ -0,0 +1,62 @@
+const loadTranslationsFile = (locale) => {
+  switch (locale) {
+    case 'en':
+      return import(/* webpackChunkName: "en-translations" */ 'en.js');
```
mjhenkes, line 6: could we use "dynamic expressions" to load any aggregated translation in the aggregate translations folder? this is similar to how you load things in the workflow server.

https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import

mjhenkes: Thinking more about this, i don't think it's nearly as easy as i want it to be due to how we're resolving the translations currently, you wouldn't sanely do this `import({locale}.js);` because it would load in way to many files (i think). To get around that we'd have to mess with our webpack config in one way or another which gets us into non passive territory pretty quick.

jmsv6d: Yeah, that was my original thinking when I first making the changes. ~~But it looks like the magic comment: `/* webpackMode: "lazy" */` might actually solve this issue.~~ JK it doesn't and lazy mode is the default setting anyway 🤦🏻‍♀️

mjhenkes (approved):

PR #164, opened 2021-03-17
Creating the ApplicationContainer, ApplicationContainerErrorBoundary, and related files

### LINKED ISSUE #149: Create ApplicationContainer component to encapsulate high-level framework concerns
# Story

## Description

As a consumer of Terra, I want to be able to use the ApplicationContainer component...

## Motivation

So that I have access to high-level framework and function in my application.

## Definition of Done

We'll know we're done when we have created the ApplicationContainer and moved relevant features out of ApplicationBase and into the ApplicationContainer.

## Details

This story will be followed by many more to add functionality to the ApplicationContainer. This story should result in a relatively empty component containing existing features of ApplicationBase that we want to remove. ApplicationBase will be primarily responsible for i18n and theming, as those are going to be universal requirements regardless of application type. The ApplicationContainer may need to flex or be replaced by other applications if we are integrating with an existing application/framework.

tbiethman:
### Summary

This PR introduces the ApplicationContainer. The ApplicationContainer will be fleshed out more in the coming iterations.

Other new stuff includes:

ApplicationContainerErrorBoundary - New error boundary specific to the ApplicationContainer. The existing ApplicationErrorBoundary will be going away with v2. We don't want to render boundaries all over the place.

ErrorBoundaryBase - ....that being said, I made a base error boundary implementation that we can reuse if we need to. I really just don't want to dupe this class boilerplate all over the place.

WindowManager - This is a pre-emptive change to get in front of our eventual need to refresh the page even while there is dirty data present (session terminations and whatnot). This singleton will allow for other components in the framework to more easily manipulate these things (for better or worse?). It'll also prevent us from getting more than one prompt if multiple handlers are bound, which might be nice at some point. 

Closes #149.

### Deployment Link

### Testing

Jest tests exhibiting 100% coverage.

### Additional Details

@cerner/terra


mjhenkes (approved):

src/application-container/private/ApplicationContainerErrorBoundary.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * The ApplicationContainerErrorBoundary provides a render error boundary for
+ * an entire application.
+ *
+ * It should not be used to provide error handling within page content.
+ */
+const ApplicationContainerErrorBoundary = ({ children }) => {
+  const applicationIntl = React.useContext(ApplicationIntlContext);
+  const [caughtError, setCaughtError] = React.useState(false);
```
yuderekyu, line 25: i'm curious, any particular reason we chain the context/state calls from the React object, instead of importing them separately first?

src/shared/ErrorBoundaryBase.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+    this.state = { error: undefined };
+  }
+
+  static getDerivedStateFromError(error) {
+    return { error };
+  }
+
+  componentDidCatch(error) {
+    logger.error(error);
```
yuderekyu, line 41: are there any scenarios where we don't want to log the error?

yuderekyu (approved):

src/application-container/private/ApplicationContainerErrorBoundary.jsx
tbiethman: There were issues in the past with mocking these individual module properties due to how module scoping works with Jest. Depending on how you test, it's not a big deal, but I got in the habit of doing it this way just in case.

src/shared/ErrorBoundaryBase.jsx
tbiethman: That's a good question. Given the use case for this error boundary (e.g. handling egregious fatal errors), I honestly don't see a scenario where we would want to suppress that error from the logs. And I wouldn't expect consumers to use this component to implement "soft" error handling (something that would typically be handled with try/catch).

src/application-container/private/ApplicationContainerErrorBoundary.jsx
yuderekyu: that's good to know!

PR #168, opened 2021-03-23
Add fallback locale incase the regional locale is unsupported

### LINKED ISSUE #143: [spike] Investigate implementing a regional locale fallback
# Feature Request

This spike/feature is to investigate adding regional locale fallback logic to the [helper utility in ApplicationBase](https://github.com/cerner/terra-application/blob/1ba7de59af383cc14a3c351ee8fbcf6762d7c233/src/application-base/ApplicationBase.jsx#L27).

## Description

The utility currently compares the preferred browser locale against the list of supported locales defined by `TERRA_AGGREGATED_LOCALES`, but does not implement any regional locale fallback logic like [terra-i18n](https://github.com/cerner/terra-core/blob/main/packages/terra-i18n/src/intlLoaders.js#L53).

To be consistent, we should discuss adding regional locale fallback logic to align with the regional locale fallback logic performed in terra-i18n.

For example, if the browser's preferred locale is `sv-SE`, but it is not included in `TERRA_AGGREGATED_LOCALES` we should check if the regional `sv` locale is supported/included in  `TERRA_AGGREGATED_LOCALES` before moving on to the next locale/check.

lokesh-0813:
### Summary

Provides logic to fallback to a locale when the regional locale is not supported.

For instance, if the browser preferred regional locale is `fr-CA` which is not supported by `TERRA_AGGREGATED_LOCALES ` this change allows to fallback to `fr` locale. 

Closes #143 

### Deployment Link

### Testing

Added jest
### Additional Details

@cerner/terra


src/application-base/private/getBrowserLocale.js
```diff
@@ -24,6 +24,13 @@ const filterLanguages = () => {
   if (navigator.languages && navigator.languages.length > 0) {
     for (let index = 0; index < navigator.languages.length; index += 1) {
       const locale = navigator.languages[index];
+      const fallbackLocale = locale.split('-').length > 1 ? locale.split('-')[0] : false;
```
StephenEsser: We should avoid mix and matching types when possible. In this case the fallbackLocale can be a `string` or a `boolean`. The exceptions are `undefined` and `null`. This assignment should always resolve to a string or `undefined`/`null`.

src/application-base/private/getBrowserLocale.js
```diff
@@ -24,6 +24,13 @@ const filterLanguages = () => {
...
+
+      if(fallbackLocale){
+        if (isSupported(fallbackLocale)) {
+          return fallbackLocale;
+        }
+      }
 
       if (isSupported(locale)) {
```
StephenEsser, line 29: We need to check the preferred locale prior to falling back to the regional locale

src/application-base/private/getBrowserLocale.js
```diff
@@ -24,6 +24,13 @@ const filterLanguages = () => {
   if (navigator.languages && navigator.languages.length > 0) {
     for (let index = 0; index < navigator.languages.length; index += 1) {
       const locale = navigator.languages[index];
+      const fallbackLocale = locale.split('-').length > 1 ? locale.split('-')[0] : false;
+
+      if(fallbackLocale){
+        if (isSupported(fallbackLocale)) {
```
StephenEsser: These logic statements can be combined instead of nested:

```js
if (fallbackLocale && isSupported(fallbackLocale))
```

src/application-base/private/getBrowserLocale.js
lokesh-0813: changed here 7731a62c36827a92ca385599cec577dc79d6fd1e

src/application-base/private/getBrowserLocale.js
lokesh-0813: changed here 7731a62c36827a92ca385599cec577dc79d6fd1e

src/application-base/private/getBrowserLocale.js
lokesh-0813: changed here 7731a62c36827a92ca385599cec577dc79d6fd1e

ShettyAkarsh (approved):

StephenEsser (approved):

mjhenkes (approved):

PR #177, opened 2021-03-30
[terra-application-v2] Upgrade to node 12

### LINKED ISSUE #175: [terra-application-v2] Node 12
# Story

## Description

As a maintainer I would like to upgrade this component to Node 12 (lts/erbium)

## Motivation

So that it can have the latest features and security updates and remain tech current with Node.

## Definition of Done

We know we're done when:

- [x] The .nvmrc file is using lts/fermium
- [x] All tests are passing
- [x] A PR is merged to the terra-application-v2 branch

mjhenkes:
### Summary

Upgrade the v2 branch to node 12.

Closes #175 

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "version": "1.35.0",
   "description": "A framework to support application development with Terra components",
   "engines": {
-    "node": ">=8.9.2 <11"
+    "node": ">=10 <13"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "version": "1.35.0",
   "description": "A framework to support application development with Terra components",
   "engines": {
-    "node": ">=8.9.2 <11"
+    "node": ">=10"
```
benbcai, line 10: Do we want this to be ">=10 <13"? Some of our projects restrict to only 10 & 12 while others allow any version >= 10. How do we determine whether or not to include `<13`? Is it based on whether or not the project is compatible with versions beyond 12?

mjhenkes: Here's my thinking:

A packages 'engines' node version should reflect the version of node that the packaged module can run on. 

We've been dropping support for the old versions of node which is fine, but i don't know of anything that would prevent the @cerner/terra-application package from running on node 14. The git repo itself won't run on node 14 because of the fibers transitive dependency but the released package doesn't have that issue.

benbcai (approved):

StephenEsser (approved):

PR #178, opened 2021-03-30
[terra-application-v2] Consume Jest config terra

### LINKED ISSUE #174: [terra-application-v2] Consume jest-config-terra
# Story

## Description

As a terra engineer I want to upgrade to the reusable testing config.

## Motivation

So that maintenance costs are reduced.

## Definition of Done

We'll know we're done when the package is included and the jest tests run successfully.

mjhenkes:
### Summary

Consume Jest config terra

Changes:
* updated recursive method to retrieve local messages
* re-worked translation loader tests. had to be changed due to jest module loader changes
* Application Navigation gets the div fix applied here in terra-framework: https://github.com/cerner/terra-framework/pull/1381
* i18n loader tests modified to not require mocks between it blocks
* mock implementations for console.error and logger.error for error boundary tests

Closes #174

### Deployment Link

### Additional Details

@cerner/terra


src/application-base/private/translationsLoaders.js
```diff
@@ -39,22 +39,26 @@ const loadTranslationsFile = (locale) => {
...
 
-const loadTranslations = (locale) => {
-  const fallbackLocale = locale.split('-').length > 1 ? locale.split('-')[0] : false;
-
-  return loadTranslationsFile(locale).catch((error) => {
-    logger.warn(`${error.message} Using ${fallbackLocale} data as the fallback locale.`);
+const getFallbackLocale = (locale) => {
+  // there is no fallback for en.
+  if (locale === 'en') {
+    return false;
```
benbcai: Should this return null or undefined instead of false to avoid a mixed return type?

mjhenkes: good idea

mjhenkes:
```suggestion
    return undefined;
```

StephenEsser (approved):

benbcai (approved):

PR #185, opened 2021-04-08
Add session context.

### LINKED ISSUE #157: [ApplicationContainer] Develop ApplicationSessionContext API
# Story

## Description

As a consumer, I want access to the ApplicationSessionContext containing session-related data...

## Motivation

So that I have a consistent way to access session APIs regardless of the actual authentication method.

## Definition of Done

We'll know we're done when we have created the ApplicationSessionContext and defined the API for its usage.

## Details

We still need to determine whether or not we want to include the concept of a "user" within this context or have something separate for that.

dkasper-was-taken:
### Summary
Introduce application session context structure.

Closes #157

### Deployment Link

@cerner/terra


src/session/SessionContext.jsx
```diff
@@ -0,0 +1,45 @@
...
+     */
+    username: PropTypes.string,
+    /**
+     * The user's first name.
+     */
+    firstName: PropTypes.string,
+    /**
+     * The user's last name.
+     */
+    lastName: PropTypes.string,
```
tbiethman, line 36: This looks great. Wondering if we want to include a separate metaData property in the use shape, or if we'd expect that to roll up into the root-level metaData object?

tbiethman (approved):

src/session/SessionContext.jsx
dkasper-was-taken: I'd be fine with either, depends on how easy it is to message the expectation to users.

tbiethman: I imagine it's fine as it is. I can see us adding more first-class keys to user over time and relying on root metaData for anything else.

src/session/SessionContext.jsx
```diff
@@ -0,0 +1,45 @@
...
+ * value is not validated, as a provider is not required for all applications.
+ * @returns The retrieved SessionContext value.
+ */
+const useSession = () => React.useContext(SessionContext);
+
+const sessionContextShape = PropTypes.shape({
+  /**
+   * Whether or not the current session is active or in an inactive state i.e. "locked".
+   */
+  isActive: PropTypes.bool,
```
mjhenkes: We may want to remove this.

dkasper-was-taken: Updated.

yuderekyu (approved):

mjhenkes (approved):

PR #187, opened 2021-04-14
Updating ApplicationBase for v2 architecture

### LINKED ISSUE #184: [terra-application-v2] Update ApplicationBase to align with new architecture
# Story

## Description

As a consumer, I want the ApplicationBase component to be able to work in conjunction with the ApplicationContainer...

## Motivation

So that I can utilize i18n/theming in my applications without defects.

## Definition of Done

We'll know we're done when ApplicationBase has been updated, tests have been updated, and examples are able to use ApplicationBase and ApplicationContainer without issue.

## Details

There are a number of things that have fallen out of date with the v1 branch, I plan on bringing those changes/fixes in with this story. The test implementations may take a hit; I might have to remove tests for components that are to be removed at a later date, or just remove them outright.

Might be able to sneak #165 in there too.

Estimated story points: 3

tbiethman:
### Summary

Updating ApplicationBase to align better with ApplicationContainer development. Introducing changes made in the v1 pipeline since branch creation.

Closes #184

### Deployment Link

### Testing

New unit tests have been written. Existing wdio tests using ApplicationBase continue to pass.

### Additional Details

@cerner/terra


src/application-base/ApplicationBase.jsx
```diff
@@ -32,97 +33,71 @@ const propTypes = {
...
   useEffect(() => {
     if (finalLocale !== undefined) {
-      i18nLoader(finalLocale).then(translationsModule => setMessages(translationsModule.default)).catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error(error);
-        throw error;
+      i18nLoader(finalLocale).then(translationsModule => {
+        setMessages(translationsModule.default);
+      }).catch((error) => {
+        logger.error(error);
```
tbiethman, line 59: I removed the thrown error here. If locales cannot be loaded, we will log and continue rendering null. Throwing an error seemed like overkill (and it was hard to test).

src/application-base/ApplicationBase.jsx
```diff
@@ -1,29 +1,30 @@
...
 import './baseStyles';
 
 const browserLocale = getBrowserLocale();
 
-// We only need to retrieve the root theme and root theme name once for the life of the application.
-const themeConfig = (typeof (TERRA_THEME_CONFIG) !== 'undefined') ? TERRA_THEME_CONFIG : undefined;
-const rootThemeName = themeConfig?.theme ? themeConfig.theme : 'terra-default-theme';
+// We only need to retrieve the root theme and root theme name once for the life
+// of the application.
+const themeConfig = (typeof (TERRA_THEME_CONFIG) !== 'undefined')
```
yuderekyu: personal preference and nitpick - i think a single line ternary is easier to understand. i'd just use a regular if statement if we wanted to break this into multiple lines 🤷🏼

src/application-base/baseStyles.js
```diff
@@ -5,6 +5,8 @@ if (process.env.NODE_ENV !== 'production') {
   // Check to ensure dir attribute is set on html element
   if (!document.documentElement.hasAttribute('dir')) {
     // eslint-disable-next-line
-    console.warn('The html element is missing the dir attribute. For terra directionality based styles to render correctly, add dir="ltr" or dir="rtl" to the html element.');
+    console.warn('The html element is missing the dir attribute. ' +
```
yuderekyu: should this be a console.error instead?

yuderekyu (approved):

src/application-base/baseStyles.js
tbiethman: Given that it can wreck certain components hard, I think it should be logged as an error. Could also go through the Logger we have vs. standard console.

src/application-base/ApplicationBase.jsx
tbiethman: Can you tell I finally turned on line rules again? I'll see if I can simplify these statements a bit.

mjhenkes (approved):

src/application-base/baseStyles.js
yuderekyu: do we want this error filtered from production? that seems to be the main difference between the Logger and standard console

tbiethman: The initial implementation of this did, at least. How about this, I'll remove the explicit env check in this file, switch to the logger that already does an env check, and keep it a warning to ensure it gets filtered out in prod. If something sneaks through to prod, an error in the console ain't gonna help much 🤷‍♂️

yuderekyu: sounds good to me!

PR #195, opened 2021-04-23
Adding MainContainer, ActiveMainContext, and related files

### LINKED ISSUE #156: [ApplicationContainer] Further develop ActiveMainPageContext and related components
# Story

## Description

As a consumer, I want access to the ActiveMainPageContext containing active page data....

## Motivation

So that I can know what the active main content of the application is at any given time.

## Definition of Done

We'll know we're done when we have created the ActiveMainPageContext and integrated it within the ApplicationContainer.

## Details

This is primarily used by the Workspace framework to determine when the main page content changes (so that it can change in turn), but this will be accessible to every component rendered within the ApplicationContainer.

Estimated Story Points: 3

tbiethman:
### Summary

Adding the MainContainer here, which provides an easy way for consumers to render accessible main content. Also added is infrastructure to communicate the applications current active main element, with integration into our navigation architecture. This is primarily used by Workspace but could have other applications in the future.

Closes #156.

### Deployment Link

### Testing

Unit and functional tests added for new functionality.

### Additional Details

@cerner/terra


src/main-container/private/ActiveMainProvider.jsx
```diff
@@ -0,0 +1,99 @@
...
+  }, []);
+
+  const activeMainRegistrationContextValue = React.useMemo(() => ({
+    register: (registrationData) => {
+      if (!registrationData) {
+        return undefined;
+      }
+
+      const { label, metaData } = registrationData;
+      const registrationId = uuidv4();
```
tbiethman, line 80: Multiple main sources can be writing to this single provider, and the order of their registrations isn't guaranteed to be deterministic. So I assign an identifier to each registration and check it above prior to performing any unregistration, ensuring that main elements can execute their unregistration logic logically as part of their lifecycle without worrying about damaging registration data from other sources.

tbiethman: This might be more deterministic in the future with React 17 based on some PRs I've seen: https://github.com/facebook/react/pull/17925

src/main-container/private/ActiveMainProvider.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+      dispatch({
+        type: 'register',
+        registrationId,
+        label,
+        metaData,
+      });
+
+      return () => {
+        dispatch({ type: 'unregister', registrationId });
```
yuderekyu, line 90: nitpick for readability/consistency
```suggestion
        dispatch({ 
          type: 'unregister', 
          registrationId,
        });
```

src/main-container/private/ActiveMainProvider.jsx
yuderekyu: we could add what you've outlined as a comment/todo

yuderekyu (approved):

benbcai (approved):

src/main-container/MainContainer.module.scss
```diff
@@ -0,0 +1,5 @@
+:local {
+  .main-container {
+    outline: none;
```
mjhenkes, line 3: should main have 100% height?

src/main-container/private/ActiveMainProvider.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+  React.useEffect(() => {
+    if (!activeMainRegistration) {
+      return;
+    }
+
+    // If an ancestor provider exists, we need to forward the active main info
+    // if the provider exists in the active navigation branch. It is
+    // otherwise unregistered, if necessary, as we do not want potentially stale
+    // information living above this provider level.
```
mjhenkes, line 50: Having nested active main providers should never be a thing right? If that happens maybe we should just blow up?

tbiethman: It is a thing, as the ActiveMainProviders are rendered both by the ApplicationContainer as well as by each NavigationItem (which will be inside the ApplicationContainer, and potentially nested even further if we have secondary/tertiary navigation concepts).

The nested providers are what allow us to limit cross-talk between different navigation branches of the application.

src/main-container/MainContainer.module.scss
tbiethman: I thought about this but erred on the side of caution. If we add a height, I feel like we'd need to make the same call for overflow handling. But that is not always desired/required (for the Page use case, or example), which opens the door to all sorts of props to disable/enable things... 

Long story long I'd like to avoid adding more styles here, for the time being at least. Makes me wonder if we should add a MainContainer to the eventual LegacyLayout and provide the default styling we used to provide there while keeping this component more generic.

mjhenkes (approved):

PR #216, opened 2021-05-21
Adding Page component and related files

### LINKED ISSUE #193: [terra-application-v2] Create Page component with APIs required for initial rendering
# Story

## Description

As a consumer of Terra, I want to use the Page component to render encapsulated content

## Motivation

So that I can utilize consistent styling, API, and composition regions within my applications.

## Definition of Done

We'll know we're done when the Page is developed with initial APIs required for presentation.

tbiethman:
### Summary

Adding the Page component, the main UI component intended for usage in the provided PageContainers.

Closes #193

### Deployment Link

### Additional Details

Theme values are not yet defined in total. A separate issue has been logged to get those implemented when the values are known: https://github.com/cerner/terra-application/issues/218

@cerner/terra


packages/terra-application/src/page/Page.jsx
```diff
@@ -0,0 +1,136 @@
...
+        </div>
+      </div>
+    </PagePortal>
+  );
+};
+
+Page.propTypes = propTypes;
+Page.Actions = PageActions;
+Page.Action = PageAction;
+Page.Toolbar = PageToolbar;
```
tbiethman, line 138: These might be a little controversial. As we have moved away from needing HOCs all over the place, the usage of static exports from some of our "umbrella" components becomes a little more straightforward and less error prone. The components themselves are documented separately so we can document prop tables and all that jazz.

I feel like this simplifies the consumer's implementation code and keeps these components (that are required for usage with the Page) bundled together nicely. Take a look at the example implementations and let me know your thoughts.

packages/terra-application/src/page/PageActions.jsx
```diff
@@ -0,0 +1,24 @@
...
+/* eslint-enable react/no-unused-prop-types */
+
+/**
+ * The PageActions component allows Page consumers to define visible actions
+ * using JSX. The Page will parse props from  PageActions and present a
+ * variety of controls for the actions depending on Page constraints.
+ *
+ * The PageActions component should not be rendered directly.
+ */
+const PageActions = () => <div />;
```
mjhenkes: it doesn't matter, but could this be an empty fragment?

packages/terra-application/src/page/PageHeaderButton.jsx
```diff
@@ -0,0 +1,78 @@
...
+   */
+  onSelect: PropTypes.func,
+  /**
+   * Ref callback for button.
+   */
+  refCallback: PropTypes.func,
+  /**
+   * Identifier for use during tests
+   */
+  testId: PropTypes.string,
```
mjhenkes: testId? Should we make this private or just be a regular id?

packages/terra-application/src/page/PageActions.jsx
tbiethman: Good question. It needs to be non-null so that the props table parsing will recognize it as a component, but I imagine a fragment would work too? I'll try it out.

tbiethman: It works!

packages/terra-application/src/page/Page.jsx
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  return (
+    <PagePortal>
+      <div className={cx('page')}>
+        <div className={cx('header')}>
+          <PageHeader
+            id={pageId}
+            onSelectBack={onRequestClose ? () => {
+              if (dangerouslyDisableUnsavedChangesPromptHandling) {
```
yuderekyu: (small and not a big deal) - we could convert this into a `handleOnSelectBack` function and pass it here to make the ternary slightly more readable

packages/terra-application/src/page/PageHeader.jsx
```diff
@@ -0,0 +1,220 @@
...
+        key={action.key}
+        refCallback={action.props.refCallback}
+        icon={action.props.icon}
+        ariaLabel={action.props.label}
+        onSelect={action.props.onSelect}
+      />
+    ));
+  }
+
+  function renderMenuButton() {
```
yuderekyu: what are your thoughts on further abstracting these render functions into their own components? Looks like we have `ActionButtons`,  `MenuButton`, `Menu`, `Toolbar`, `StartActions`, `EndActions`. Extracting these out will remove nested rendering and make this component easier to follow

packages/terra-application/src/page/PageHeader.jsx
```diff
@@ -0,0 +1,220 @@
...
+      />
+    );
+  }
+
+  function renderMenu() {
+    return (
+      <Popup
+        isOpen={showMenu}
+        targetRef={() => moreActionsButtonRef.current}
+        onRequestClose={() => { setShowMenu(false); }}
```
yuderekyu: nitpick haha 
```suggestion
        onRequestClose={() => setShowMenu(false)}
```

packages/terra-application/src/page/PageHeader.jsx
```diff
@@ -0,0 +1,220 @@
...
+        contentAttachment="top right"
+        isContentFocusDisabled
+        isHeaderDisabled
+        popupContentRole="none"
+      >
+        <ActionMenu
+          label={intl.formatMessage({
+            id: 'terraApplication.pageHeader.actionsMenu',
+          }, { label })}
+          onRequestClose={() => { setShowMenu(false); }}
```
yuderekyu: if you update above, let's do the same here for consistency
```suggestion
          onRequestClose={() => setShowMenu(false)}
```

packages/terra-application/src/page/Page.jsx
```diff
@@ -0,0 +1,136 @@
...
+   * provided, the back button will not be rendered.
+   */
+  onRequestClose: PropTypes.func,
+  /**
+   * When true, the Page will not prompt the user prior to executing
+   * `onRequestClose` in the presence of rendered UnsavedChangesPrompts.
+   * Use this prop to customize UnsavedChangesPrompt handling prior to Page
+   * closure.
+   */
+  dangerouslyDisableUnsavedChangesPromptHandling: PropTypes.bool,
```
mjhenkes, line 53: remind me, in what situations does this get used?

tbiethman: This is the "I'm handling my Page's unsaved changes differently and don't want two prompts to pop up" prop. I'd anticipate something similar on the new Modal.

packages/terra-application/src/page/PageHeader.jsx
tbiethman: Separate components would result in a lot of props being passed around and seems like overkill (for what would amount to a single-use component). But I might be able to make this more straightforward, I'll play with it a bit.

packages/terra-application/src/page/Page.jsx
mjhenkes: I like the explicit bundling of the page and it's sub components. My only concern is introducing inconsistency regarding how we present other API's in the package, such as workspace. It may be worth our time to go back and implement this style of api for those other areas at least in the terra-application package.

packages/terra-application/src/page/PageHeaderButton.jsx
tbiethman: I don't think this is being used at the moment, so I'll remove. Should be able to query in tests based on role/display just fine.

packages/terra-application/src/page/Page.jsx
tbiethman: Agreed, I can log an investigative story for that as part of our v2 release. We could remain passive by keeping any existing explicit exports in place.

yuderekyu (approved):

mjhenkes (approved):

packages/terra-application/src/page/Page.jsx
tbiethman: Logged issue: https://github.com/cerner/terra-application/issues/219

PR #264, opened 2021-09-01
[terra-dev-site] Fix redirect script path error

mjhenkes:
### Summary

I was not considering the public path when adding the redirect script to head, thus it wasn't being found when deployed to gh-pages. This fix corrects that.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


pranav300 (approved):

ryanthemanuel (approved):

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -236,18 +236,23 @@ class SitePlugin {
...
-    // Use default public path else the env else /
-    let defaultPublicPath;
-    if (compiler.options.output && compiler.options.output.publicPath) {
-      defaultPublicPath = compiler.options.output.publicPath;
+    if (compiler.options.output && compiler.options.output.publicPath && compiler.options.output.publicPath !== 'auto') {
+      return compiler.options.output.publicPath;
     }
-    const publicPath = process.env.TERRA_DEV_SITE_PUBLIC_PATH || defaultPublicPath || '/';
+    compiler.options.output.publicPath = '/';
+    return '/';
```
pranav300: Can we just return `compiler.options.output.publicPath`. Or was it intentional to return `/`.

benbcai (approved):

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
mjhenkes: in this case we either don't have a default or it's set to auto so we want to explicitly set it to / for our routing to work.

PR #270, opened 2021-10-04
[terra-dev-site] Validate relative urls when building the site.

mjhenkes:
### Summary

With this PR we can validate relative urls specified in md or mdx files against the relative urls generated as part of the site.

We can pull data from two different loaders, the markdown loader, and the bootstrap loader, then compare the outputs during the 'done' compiler hook.

There is currently not a way to ignore links, but it is a good idea for the future.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-application-docs/src/terra-dev-site/tool/terra-dev-site/UpgradeGuides.f/v5.0.0.u.tool.mdx
```diff
@@ -36,8 +36,8 @@ Reference [terra-toolkit's v4 upgrade guide](https://github.com/cerner/terra-too
 
 Terra dev site now requires peer dependencies on the following modules:
 
-- terra-toolkit@5.0.0
-- webpack@4.28.1
+- `terra-toolkit@5.0.0`
```
mjhenkes, line 39: these were showing up as 'mailto' links

packages/terra-application-docs/src/terra-dev-site/tool/terra-dev-site/about.a.tool.mdx
```diff
@@ -127,7 +127,7 @@ Without the environment variable set assets will assume they are served from `/`
 
 TerraDevSite is built to work out of the box but it also provides many configuration options to allow you to customize your site.
 
-Check out our [configuration docs](/dev_tools/terra-dev-site/terra-dev-site/configuration).
+Check out our [configuration docs](/dev_tools/cerner-terra-application-docs/terra-dev-site/configuration).
```
mjhenkes, line 130: Bad link

packages/terra-dev-site/src/webpack/loaders/devSiteBootstrap.js
```diff
@@ -58,6 +59,15 @@ const loader = async function loader(template) {
     logger: this.getLogger ? this.getLogger('terra-dev-site loader') : undefined,
   });
 
+  if (dataCallback) {
```
mjhenkes, line 62: this call back send the generated links back to the main plugin for later use.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -3,19 +3,27 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const path = require('path');
 const { DefinePlugin } = require('webpack');
 const fs = require('fs');
+const logging = require('webpack/lib/logging/runtime');
+const chalk = require('chalk');
+// const https = require('https');
```
mjhenkes, line 8: these are left here on purpose. see my comment farther down for explanation

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -341,6 +359,72 @@ class SitePlugin {
...
+
+        if (url.url.startsWith('/')) {
+          // for relative urls check against the know list of valid relative urls.
+          // Remove query params or hash links from the end of the url.
+          const urlWithoutHash = url.url.replace(/[#?].*$/, '');
+          if (!validRelativeUrls.includes(urlWithoutHash)) {
+            resultPromises.push(Promise.resolve(`${chalk.red('[Broken Link]')} ${invalidUrlLog(url)}`));
+          }
+        } else if (url.url.startsWith('https')) {
+          // required to not assume https urls are invalid.
```
mjhenkes, line 387: so we could validate more link types, http, https and even maybe ./ style relative links, but there would need to be some more changes. 

For http and https style links, the real trouble is that link to engineering.cerner.com/terra-ui always come back as 404s and we don't currently have a way to check if the links are valid. That the links come back as 404s is not unexpected because we use a custom 404 to enable the browser router and to redirect to the correct page. We would need to look into further identification of what is a bad link in a deployed site for this to work out.

for ./ style relative links we'd need to generate the site path based on the file that is hosting the link. Doable but im not sure it's worth it.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -341,6 +359,72 @@ class SitePlugin {
...
+          //       resolve();
+          //     }
+          //   });
+          // }));
+        } else if (url.url.startsWith('#')) {
+          // Assume anchors are fine.
+          // We could check the document for an id to ensure the anchor link will take you somewhere but thats beyond the initial scope.
+        } else if (url.url.startsWith('.')) {
+          // To validate these we'd have to know the url for the current document. doable, but may need some duplicate logic.
+        } else {
```
mjhenkes, line 414: there is an assumption here that mailtos are not normal, although we do use a mailto correctly in terra-ui. So that would always warn... we could change this to simply check relative links only to avoid any false positives.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -341,6 +359,72 @@ class SitePlugin {
...
+          // resultPromises.push(new Promise((resolve) => {
+          //   http.get(url.url, (response) => {
+          //     if (response.statusCode >= 400) {
+          //       resolve(`${chalk.red('[Broken Link]')} ${chalk.blue(url.url)} in:\n ${chalk.yellow(url.file.history[0].replace(url.file.cwd, ''))} at line ${url.node.position.start.line}`);
+          //     } else {
+          //       resolve();
+          //     }
+          //   });
+          // }));
+        } else if (url.url.startsWith('#')) {
```
mjhenkes, line 409: we could validate here that the id is on the page, haven't done that yet.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -341,6 +359,72 @@ class SitePlugin {
...
+      const queueLength = urlQueue.length;
+      for (let index = 0; index < queueLength; index += 1) {
+        const url = urlQueue.shift();
+
+        if (url.url.startsWith('/')) {
+          // for relative urls check against the know list of valid relative urls.
+          // Remove query params or hash links from the end of the url.
+          const urlWithoutHash = url.url.replace(/[#?].*$/, '');
+          if (!validRelativeUrls.includes(urlWithoutHash)) {
+            resultPromises.push(Promise.resolve(`${chalk.red('[Broken Link]')} ${invalidUrlLog(url)}`));
```
mjhenkes, line 384: this is returned with a promise to enable http or https checking in the future that must be asynchronous. this has the added benefit of not slowing down webpack.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
```diff
@@ -341,6 +359,72 @@ class SitePlugin {
...
+          //       resolve( `${chalk.red('[Broken Link]')} ${chalk.blue(url.url)} in:\n ${chalk.yellow(url.file.history[0].replace(url.file.cwd, ''))} at line ${url.node.position.start.line}`);
+          //     } else {
+          //       resolve();
+          //     }
+          //   });
+          // }));
+        } else if (url.url.startsWith('http')) {
+          // required to not assume http urls are invalid.
+          // resultPromises.push(new Promise((resolve) => {
+          //   http.get(url.url, (response) => {
```
mjhenkes, line 401: it's really annoying that https can't handle http calls or that they're separate at all... thanks node.

packages/terra-dev-site/src/webpack/plugin/siteLoaderUtils.js
```diff
@@ -1,5 +1,6 @@
 const rehypeSlug = require('rehype-slug');
 const rehypeUrl = require('rehype-urls');
+const rehypeUrlInspector = require('@jsdevtools/rehype-url-inspector');
```
mjhenkes, line 3: we use this plugin because it also gives you back the file and location of the link.

packages/terra-dev-site/src/webpack/plugin/SitePlugin.js
benbcai: Or we could use a light-weight third-party lib but it's nice to just use the native since it's already available.

benbcai (approved):

ryanthemanuel (approved):

PR #277, opened 2021-11-07
Adding activation indicators and APIs to NavigationItem and Page components

tbiethman:
### Summary

The navigation and page concepts utilize portals to keep their contents mounted, even when those contents are not visible. Because of this portal logic, a component's element refs may not yet be present on the DOM during the normal component lifecycle. 

Here we introduce an API through the NavigationItemContext/PageContext that indicates whether a consumer of those contexts is "active" or not; when active, the component is guaranteed to be present on the DOM.


ryanthemanuel (approved):

neilpfeiffer (commented):
ApplicationV2Scope.5.app.mdx -> +1

Issue #280, opened 2021-11-17
[terra-dev-site] Fails to generate site on windows devices due to windows style file paths


revoyrm:
# Bug Report

## Description
When developing on windows and running npm start, it outputs errors caused by windows file paths with backslashes instead of forward slashes. The errors are in terra-dev-site\webpack\loaderUtils\generateNavigationConfig.js and can be fixed in executeSearchPatterns and generateNavigationConfig. Below are the 2 errors.

Error 1: 
![image](https://user-images.githubusercontent.com/11396904/142272736-d15e7ed8-eb17-42fa-b72b-188483011aef.png)

When I fix that error in node_modules/@cerner/terra-dev-site/webpack/loaderUtils/generateNavigationConfig.js - executeSearchPatterns, I get the 2nd windows file path error.

Error 2: 
![image](https://user-images.githubusercontent.com/11396904/142273145-a7906460-3efb-4049-8b16-20fad9628b28.png)

Which can be fixed in  terra-dev-site/webpack/loaderUtils/generateNavigationConfig.js - generatePagesConfig.

## Steps to Reproduce

1. Run a site using terra-dev-site 7.0.0 on a windows device

## Additional Context / Screenshots
Here is a link to a conversation in teams about this issue 
https://teams.microsoft.com/l/message/19:3ed4d4d6ffba48138550b9a91a2b50b4@thread.skype/1634129949052?tenantId=fbc493a8-0d24-4454-a815-f4ca58e8c09d&groupId=45f47bfa-38a9-444d-8933-9c20266ae38c&parentMessageId=1634129949052&teamName=Terra&channelName=UX%20Support&createdTime=1634129949052

## Possible Solution
Below is the fix which I've tested on windows and WSL. I have not been able to test it on a mac as I do not have access to one. 

The fix is to run .replace(/\\\\/g, '/') on the following 3 variables:

Error 1: 
entryPoint and filePath on line -> https://github.com/cerner/terra-application/blob/main/packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js#L239

Error 2: 
content.filepath on this line -> https://github.com/cerner/terra-application/blob/main/packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js#L305

This replace "\\\\" with "/". It should only affect windows file paths and leave linux style file paths alone.

## Environment

* Component Name and Version: EDM/media-viewer-js
* Browser Name and Version: Chrome
[...]

PR #281, opened 2021-11-19
UXPLATFORM-6038 - generateNavigationConfig converts windows file paths to linux style file paths

revoyrm:
### Summary

Changes required to generate development site on windows devices.

UXPLATFORM-6038 and see Issue #280 

### Testing

I ran npm run compile and moved the generated lib folder into the node modules terra dev site folder in my media-viewer-js project and ran npm start in windows and in wsl to make sure this worked in both places.

If there's a way to deploy this change so I can get it via an npm install, I will test it that way.

### Additional Details

My changes remove the double backslash in the windows file paths and replaces each instance with one forward slash. This works on both windows and linux development environments.

@cerner/terra

### DIFF
--- a/packages/terra-dev-site/src/webpack/loaderUtils/configHelpers.js
+++ b/packages/terra-dev-site/src/webpack/loaderUtils/configHelpers.js
@@ -80,12 +80,20 @@ const parseExtension = filePath => {
   });
 };
 
+/**
+ * Returns a linux style file path.
+ *
+* @param {*} filePath A string representing the directory path of the file.
+*/
+const formatPath = filePath => filePath.replace(/\\/g, '/');
+
 const configHelpers = {
   startCase,
   pageTypes,
   getNamespace,
   getRoutes,
   parseExtension,
+  formatPath,
 };
 
 module.exports = configHelpers;
--- a/packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js
+++ b/packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js
@@ -7,6 +7,7 @@ const {
   getNamespace,
   getRoutes,
   parseExtension,
+  formatPath,
 } = require('./configHelpers');
 
 /**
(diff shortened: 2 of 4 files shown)


revoyrm:
@benbcai It appears to have failed trying to log in https://app.travis-ci.com/github/cerner/terra-application/jobs/549431493

It appears I do not have permissions to retrigger the build.

packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js
```diff
@@ -302,7 +303,7 @@ const generatePagesConfig = ({
     if (item.additionalContent) {
       item.additionalContent.forEach((content) => {
         filePaths.push({
-          filePath: content.filePath,
+          filePath: formatPath(content.filePath),
           entryPoint: `/${content.label}.${item.contentExtension}${path.extname(content.filePath)}`,
```
benbcai, line 307: Do we also need to call formatPath on this `content.filePath`?

packages/terra-dev-site/src/webpack/loaderUtils/configHelpers.js
```diff
@@ -80,12 +80,20 @@ const parseExtension = filePath => {
   });
 };
 
+/**
+ * Returns a linux style file path.
+ *
+* @param {*} filePath A string representing the directory path of the file.
+*/
+const formatPath = filePath => filePath.replace(/\\/g, '/');
```
benbcai, line 88: I was thinking we should replace with [path.sep](https://nodejs.org/api/path.html#pathsep) but I think it's ok to replace with a hard-coded `/` since that is what we're using in most of the paths in generateNavigationConfig.

packages/terra-dev-site/src/webpack/loaderUtils/generateNavigationConfig.js
revoyrm: We don't need it on line 307 because path.extname returns the extension correctly either way. It is only needed for the filePath property on line 306.

benbcai (approved):

BenBoersma (approved):

jmsv6d (approved):

benbcai:
Replaced in #284

PR #289, opened 2022-02-02
Add support for Node 14.

benbcai:
### Summary

- Add support for Node 14.
- Upgrade to terra-functional-testing to V2.

Closes #140 

### Deployment Link

https://terra-applic-.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #291, opened 2022-02-14
Make node version open-ended

benbcai:
### Summary

- PR #289 close-ended the Node version up to 14 and this affected consumers on 16+.
- Reverted the Node version to support versions greater than 14.
- Added doc for Node version support. 
- Publish these packages

Closes #

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-dev-site/README.md
+++ b/packages/terra-dev-site/README.md
@@ -26,6 +26,10 @@ npm install --save-dev @cerner/terra-dev-site
 npm install --save-dev terra-application react-dom@^16.8.5 react@^16.8.5 webpack@^5
 ```
 
+## Node version support
+
+This package was developed and tested using Node 10 up to Node 14. Consumers using Node 16 or greater are advised to use it at their own risk since those versions are not officially supported due to lack of thorough testing.
+
 ## Versioning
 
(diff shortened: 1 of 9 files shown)


packages/terra-dev-site/README.md
```diff
@@ -26,6 +26,10 @@ npm install --save-dev @cerner/terra-dev-site
 npm install --save-dev terra-application react-dom@^16.8.5 react@^16.8.5 webpack@^5
 ```
 
+## Node version support
+
+This package was developed and tested using Node 10 up to Node 14. Consumers using Node 16 or greater are advised to use it at their own risk since those versions are not officially supported due to lack of thorough testing.
```
xenoworf, line 31: what about Node 11, and 13 - did anyone test that? If no, just change the doc to call out the specific versions we do know we support.

sdadn (approved):

packages/terra-dev-site/README.md
benbcai: Odd Node versions are non-LTS versions...more like beta. So we only use even (LTS) versions.

xenoworf: Thanks for the knowledge!

xenoworf (approved):

PR #304, opened 2022-06-24
Delete RELEASE.md

xenoworf:
I want to unify our release instructions across these mono repos. It's less maintenance and realistically no one from the open source community is ever going to cut a release of this.


supreethmr (approved):

supreethmr:
Instead of deleting the release.md can we update it to provide information about wiki page where the release steps are updated now. so that new members in team will get to know where to look for release steps !!

xenoworf:
> Instead of deleting the release.md can we update it to provide information about wiki page where the release steps are updated now. so that new members in team will get to know where to look for release steps !!

I would prefer not to link multiple Terra repos to one internal wiki page, so instead I will link the release steps prominently in a good spot and announce it in our internal chat.

PR #316, opened 2022-11-16
Fixes Scroll and Cliping issues of dev-site v7

supreethmr:
### Summary

This PR Patchs Back the code changes made on PR #310 https://github.com/cerner/terra-application/pull/310 for fixing height issue with dev-site v7.

Page with longer content was clipped to height of card ( container of dev-site content ) 

Screen Shot 2022-11-17 at 4 30 44 PM

flex box styles have been added for dev-site wrappers to allow content to grow without getting clipped. 

Closes #UXPLATFORM-6906

### Deployment Link

https://terra-applic-devsite-v7-1iourz.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #318, opened 2022-12-19
[terra-dev-site] fix card scrolling in terra-dev-site

sambao21:
### Summary
The card container that displays content wasn't scrolling for me, and by setting terra-dev-site card overflow from hidden to 

### Testing
Loaded up the terra-dev-site in Chrome and Firefox before and after the fix to show that it was not scrolling before, and it does after changing overflow to auto.

Before:
Chrome:
Kapture 2022-12-19 at 16 05 35

Firefox:
Kapture 2022-12-19 at 16 06 49

After:
Chrome:
Kapture 2022-12-19 at 16 02 08

Firefox:
Kapture 2022-12-19 at 16 04 07

### DIFF
--- a/packages/terra-dev-site/src/pages/page/layouts/Card.module.scss
+++ b/packages/terra-dev-site/src/pages/page/layouts/Card.module.scss
@@ -16,7 +16,7 @@
     flex-direction: column;
     height: calc(100% - 30px);
     min-height: calc(100% - 30px);
-    overflow: hidden;
+    overflow: auto;
     padding: 0 10px 10px;
 
     @include terra-mq-medium-up {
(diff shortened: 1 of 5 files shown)


packages/terra-dev-site/src/pages/page/layouts/Card.module.scss
```diff
@@ -16,7 +16,7 @@
     flex-direction: column;
     height: calc(100% - 30px);
     min-height: calc(100% - 30px);
-    overflow: hidden;
+    overflow: auto;
```
neilpfeiffer: Thanks @sambao21 - apologies, we actually resolved the issue already with one of our recent PRs #316, which fixed the issues with both clipping and scrolling. But in the rush to get it out for the consumers needing it patched-back for dev-site@v7, we then forgot to apply the changes also to main for the latest and dev-site@v8.  

If you could apply the same changes from #316, that would be the design approach we would like to resolve the issue -- where the it is the page, with a content-sized card on it, overflows and scrolls. Not allowing the scrolling to be on the interior of the card as this change would make.

supreethmr (approved):

BenBoersma (approved):

supreethmr:
Since the similar changes were already approved by UX on PR https://github.com/cerner/terra-application/pull/316 

Merging this one to fix dev-site issue on dev-site v8

PR #327, opened 2023-03-24
Add ability to use Docker Dev containers locally

nikhitasharma:
### Summary

Add ability to use Docker Dev Container locally
Update Readme

@cerner/terra

### DIFF
--- a/.devcontainer/docker-compose-wdio.yml
+++ b/.devcontainer/docker-compose-wdio.yml
@@ -1,0 +1,43 @@
+version: "3"
+
+services:
+  selenium-hub:
+    image: selenium/hub:3.141.59-zirconium
+    container_name: selenium-hub
+    ports:
+      - "4444:4444"
+    healthcheck:
+      test: ["CMD", "/opt/bin/check-grid.sh"]
+      interval: 15s
+      timeout: 30s
+      retries: 5
+
+  chrome:
+    image: selenium/node-chrome:3.141.59-zirconium
+    # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
+    # For more details see:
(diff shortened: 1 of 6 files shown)


trandrew1023 (approved):

.devcontainer/docker-compose-wdio.yml
```diff
@@ -0,0 +1,43 @@
...
+    #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
+    #  https://bugs.chromium.org/p/chromium/issues/detail?id=519952
+    #  https://bugzilla.mozilla.org/show_bug.cgi?id=1338771#c10
+    volumes:
+      - /dev/shm:/dev/shm
+    depends_on:
+      - selenium-hub
+    environment:
+      - HUB_HOST=selenium-hub
+      - HUB_PORT=4444
```
seanturk, line 43: Should've mentioned this on the last one - do we need an Edge config?

nikhitasharma: Right now, none of the terra WDIO tests use Edge. I added this to match with Terra CICD. 
If you think we will be adding Edge WDIO tests in the near future, then we would need to update the devcontainer as well as the terra-functional-testing docker compose file which gets pulled during pipeline build execution from here: https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/docker/docker-compose.yml

seanturk: Might be worth an evaluation at some point since Edge is going to end up being the most used browser for Terra once all is said and done, but that's not a problem we can solve right now.

seanturk (approved):

PR #355, opened 2023-09-07
[Application-Base] Added prop to remove scroll from powerchart

saket2403:
### Summary

Added prop to remove scroll from powerchart.
**What was changed:**
`noScroll` prop was added to remove the scroll by adding a custom style to terra-scroll.

**Why it was changed:**
An additional scroll appeared in powerchart for viewpoint tab list control when it was wrapped with App base.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] Visual testing (please attach a screenshot or recording)
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

In addition to engineering reviews, this PR needs:

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9568 

---

@cerner/terra


sugan2416 (approved):

packages/terra-application/src/application-base/ApplicationBase.jsx
```diff
@@ -76,10 +77,16 @@ const propTypes = {
    * before continuing with the unload event, even if NavigationPrompts are present.
    */
   unloadPromptIsDisabled: PropTypes.bool,
+  /**
+   * @private
+   * Prop to not render scroll in  Powerchart.
```
supreethmr:
```suggestion
   * When set to true scroll will be disabled. internal prop to be used by Mpages for terra-tabs in  Powerchart.
```

packages/terra-application/src/application-status-overlay/ApplicationStatusOverlayProvider.module.scss
```diff
@@ -23,4 +23,8 @@
     width: 100%;
     z-index: 1;
   }
+
+  .remove-scroll {
+    overflow: visible;
```
supreethmr, line 28: https://caniuse.com/?search=overflow%3A%20visible 
overflow-visible is not supported on IE 10 and IE 11 have we tested this on the IE11 and IE10 browsers.

saket2403: Since this is an MPage-specific change and MPages 9.0 supports only edge. We do not have to worry about IE.

supreethmr (approved):

PR #366, opened 2024-01-23
Updated redwood theme's default density to 'comfortable'

artpark:
### Summary

**What was changed:**

The default density for redwood has been updated to 'comfortable'.

**Why it was changed:**

Consumers could forget the density option, this will ensure something still displays even if they forget it.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] Visual testing (please attach a screenshot or recording)
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

In addition to engineering reviews, this PR needs:

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra

### DIFF
--- a/packages/terra-application/src/application-base/ApplicationBase.jsx
+++ b/packages/terra-application/src/application-base/ApplicationBase.jsx
@@ -123,7 +123,7 @@ const ApplicationBase = ({
   const { localeOverride, themeOverride } = useTestOverrides(); // Allows us to test deployed applications in different locales.
 
   let density = themeDensity || themeConfig?.density;
-  if (!density && themeName === 'redwood-theme') {
+  if (themeName === 'redwood-theme' && !density) {
     density = 'comfortable';
   }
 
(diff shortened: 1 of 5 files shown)


adoroshk (approved):

packages/terra-application/src/application-base/ApplicationBase.jsx
```diff
@@ -122,12 +122,17 @@ const ApplicationBase = ({
 
   const { localeOverride, themeOverride } = useTestOverrides(); // Allows us to test deployed applications in different locales.
 
+  let density = themeDensity || themeConfig?.density;
+  if (!density && themeName === 'redwood-theme') {
```
sycombs: Small suggestion: If we swap the conditions
```suggestion
  if (themeName === 'redwood-theme' && !density) {
```
then it will only check density when the theme is redwood since it will short-circuit and only the first condition needs to be evaluated.

sycombs (approved):

